PR #2, opened 2020-12-11
Completed DOM lab

j-armah:


README.md
```diff
@@ -71,6 +71,9 @@ Using the `traveler` object, update the DOM to show the traveler's *name*, *nick
 
 **YOUR NOTES**
 ```
+Can acces data.js because it's in same file directory, or maybe because of
+<script src="data.js"></script> in index.html? Not sure
```
ihollander, line 75: It's because of the <script> tag! Any Javascript files that we want loaded in the browser must be added with a <script> tag.

index.js
```diff
@@ -3,16 +3,51 @@ const header = document.querySelector("#header")
...
+    // Create elements
+
+    let li = document.createElement("li")
+    let p = document.createElement("p")
+    let img = document.createElement("img")
+    let a  = document.createElement("a")
+
+    // Add attributes and properties
 
+    li.setAttribute("data-id", sighting.id)
```
ihollander, line 35: For dataset attributes, you can also use this syntax:

```js
li.dataset.id = sighting.id
```

```html
<li data-id="1">...</li>
```

index.js
```diff
@@ -3,16 +3,51 @@ const header = document.querySelector("#header")
...
+
+    let listing = document.querySelector("#animals")
+
+    li.append(p, img, a)
+    listing.append(li)
+})
 
 /***** Deliverable 5 *****/
+
+findSighting = document.querySelector("li[data-id='3']")
```
ihollander, line 52: Nice work on the query selector! Remember to always create variables with `let` or `const` tho:

```js
let findSighting = document.querySelector("li[data-id='3']")
```

ihollander:
Nice work!

PR #3, opened 2020-12-11
done

escohen115:


index.html
```diff
@@ -5,8 +5,8 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
-  <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="data.js" defer></script>
```

index.js
```diff
@@ -5,14 +5,39 @@ console.log("Here's your header:", header)
...
 
 
+let ul = document.querySelector("#animals")
+traveler.animalSightings.forEach(function (sighting){
+    ul.innerHTML += `
+    <li data-id="${sighting.id}">
+        <p>${sighting.description}</p>
+        <img src="${sighting.photo}" alt="${sighting.species}"/>
+    <a href="${sighting.link}" target="_blank">Here's a video about the species!</a>
+    </li>`
```
ihollander, line 34: Nice work on this! Just be careful of using `innerHTML +=` since it _destroys_ any existing DOM elements inside the element. This can cause have unintended consequences when we start adding event listeners to DOM elements.

ihollander:
Great work!

PR #4, opened 2020-12-11
JS Dom mini challenge

wickedvic:
finished 1-3. got stuck on 4

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,13 +5,40 @@ console.log("Here's your header:", header)
 
 /***** Deliverable 2 *****/
 
-
+header.style.color = "teal"
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+
+const img = document.querySelector(".traveler img")
+const h2 = document.querySelector(".traveler h2")
+const em = document.querySelector(".traveler em")
+
+img.src = traveler.photo
+img.alt = traveler.name
+h2.textContent = traveler.name
+em.textContent = traveler.nickname
+
 
 
 /***** Deliverable 4 *****/
+
+
+
+const li = document.createElement("li")
+const img = document.createElement("img")
+const p = document.createElement("p")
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -5,14 +5,41 @@ console.log("Here's your header:", header)
...
-
+header.style.color = "teal"
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+
+const img = document.querySelector(".traveler img")
+const h2 = document.querySelector(".traveler h2")
+const em = document.querySelector(".traveler em")
```
ihollander, line 15: (Edit - I just re-read the HTML and saw the traveler class on the element above profile -- disregard the last comment!)

index.js
```diff
@@ -5,14 +5,41 @@ console.log("Here's your header:", header)
...
+const a = document.createElement("a")
+
+li.dataset.id = traveler.id
+p.textContent = traveler.description
+img.src = traveler.photo
+img.alt = traveler.species
+a.href = traveler.link
+
+li.append(p, anchor)
+/// i got stuck at this point
```
ihollander, line 40: You're on the right track in terms of element creation, but you should be creating elements for each of the _animalSightings_, not for the individual _traveler_. Here's a solution using `forEach`:

```js
const sightingsUl = document.querySelector("#animals");
traveler.animalSightings.forEach(function(sighting) {
  const li = document.createElement("li");
  const img = document.createElement("img");
  const p = document.createElement("p");
  const a = document.createElement("a");

  li.dataset.id = sighting.id;
  p.textContent = sighting.description;
  img.src = sighting.photo;
  img.alt = sighting.species;
  a.href = sighting.link;
  a.textContent = `Here's a video about the ${sighting.species} species!`;

  li.append(p, img, a);

  sightingsUl.append(li);
});
```

ihollander:
I added a few comments - make sure you're testing your code as you go, and make sure your query selectors are working! Overall though, you're on the right track 👍

PR #5, opened 2020-12-11
Done

bstizzle:


index.js
```diff
@@ -1,18 +1,44 @@
...
 /***** Deliverable 4 *****/
+function createAnimalLi(animal){ 
+    let ul = document.querySelector("#animals");
+
+    let li = document.createElement("li");
+    let p = document.createElement("p");
+    let img = document.createElement("img");
+    let a = document.createElement("a");
+
+    li.setAttribute('data-id', animal.id);
```
ihollander, line 27: For dataset attributes, you can also use this syntax:

```js
li.dataset.id = sighting.id
```

```html
<li data-id="1">...</li>
```

index.js
```diff
@@ -1,18 +1,44 @@
...
+    p.innerText = animal.description;
+    img.src = animal.photo;
+    img.alt = animal.species;
+    a.href = animal.link;
+    a.innerText = `Here's a video about the ${animal.species} species!`;
+
+    li.appendChild(p);
+    li.appendChild(img);
+    li.appendChild(a);
```
ihollander, line 36: You can also use `append` instead of `appendChild` to add multiple elements in one go:

```js
li.append(p, img, a);
```

index.js
```diff
@@ -1,18 +1,44 @@
...
+    ul.appendChild(li);
+};
 
+traveler.animalSightings.forEach(createAnimalLi);
```
ihollander, line 40: Great work with this callback syntax 👍

ihollander:
Excellent work!

PR #6, opened 2020-12-11
done

ConnorMul:


index.html
```diff
@@ -5,8 +5,8 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
-  <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="data.js" defer></script>
```

index.js
```diff
@@ -1,18 +1,51 @@
...
+
+travelerName.textContent = traveler.name
+travelerNickName.textContent = traveler.nickname
+travelerImg.src = traveler.photo
+travelerImg.alt = traveler.name
 
 
 /***** Deliverable 4 *****/
 
+traveler.animalSightings.forEach( function(obj) {
```
ihollander, line 27: It would be better to use a more descriptive variable name than `obj` here - like `animalObj` or `sightingObj`.

index.js
```diff
@@ -1,18 +1,51 @@
...
+    let li = document.createElement("li")
+    let p = document.createElement("p")
+    let img = document.createElement("img")
+    let a = document.createElement("a")
+
+    li.dataset.id = `${obj.id}`
```
ihollander, line 33: Using string interpolation here isn't necessary -- you can just assign the number directly:

```js
li.dataset.id = obj.id
```

index.js
```diff
@@ -1,18 +1,51 @@
...
+
+    li.append(p, img, a)
+    ul.append(li)
+})
 
 
 /***** Deliverable 5 *****/
+
+
+let dataToDelete = document.querySelector("[data-id='3']")
```
ihollander, line 50: Nice work on this query selector!

ihollander:
Excellent work!

PR #7, opened 2020-12-11
Done

antrubbo:
Felt way better about this one today than yesterday's!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,18 +1,50 @@
 /***** Deliverable 1 *****/
-const header = document.querySelector("#header")
-console.log("Here's your header:", header)
+let header = document.querySelector("h1#header")
+console.log("Here's your header:", header.textContent)
 
 
 /***** Deliverable 2 *****/
-
+header.style.color = "red"
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+
+let travelerSpecs = document.querySelector("#profile")
+h2ToUpdate = travelerSpecs.querySelector("h2")
+h2ToUpdate.textContent = traveler.name
+
+emToUpdate = travelerSpecs.querySelector("em")
+emToUpdate.textContent = traveler.nickname
+
+imgToUpdate = travelerSpecs.querySelector("img")
+imgToUpdate.src = traveler.photo
+imgToUpdate.alt = traveler.name
 
 
 /***** Deliverable 4 *****/
 
-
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1,18 +1,50 @@
...
+h2ToUpdate = travelerSpecs.querySelector("h2")
+h2ToUpdate.textContent = traveler.name
 
-/***** Deliverable 4 *****/
+emToUpdate = travelerSpecs.querySelector("em")
+emToUpdate.textContent = traveler.nickname
+
+imgToUpdate = travelerSpecs.querySelector("img")
+imgToUpdate.src = traveler.photo
+imgToUpdate.alt = traveler.name
```
ihollander, line 22: Make sure to always use `let` or `const` when declaring variables -- you did it correctly for `travelerSpecs`, but your variables for `h2ToUpdate`, `emToUpdate`, and `imgToUpdate` also need `let` when you declare the variable.

index.js
```diff
@@ -1,18 +1,50 @@
...
 
+/***** Deliverable 4 *****/
+
+traveler.animalSightings.forEach( function(animalSighting) {
+    let li = document.createElement('li')
+    let p = document.createElement('p')
+    let img = document.createElement('img')
+    let a = document.createElement('a')
+    //
+    li.id = animalSighting.id
```
ihollander, line 33: For this element, the goal is to set the `data-id` attribute, not the `id` attribute. You can set the `data-id` attribute like this:

```js
    li.dataset.id = animalSighting.id 
```

That will create an element like this:

```html
<li data-id="1">
```

index.js
```diff
@@ -1,18 +1,50 @@
...
+    img.src = animalSighting.photo
+    img.alt = animalSighting.species 
+    a.href = animalSighting.link
+    a.target = "_blank"
+    a.textContent = `Here's a video about the ${animalSighting.species} species!`
+    // 
+    let animals = document.querySelector("#animals")
+    li.append(p)
+    li.append(img)
+    li.append(a)
```
ihollander, line 43: With `.append`, you can append multiple elements at once:

```js
li.append(p, img, a)
```

index.js
```diff
@@ -1,18 +1,50 @@
...
+    animals.append(li)
+})
 
 /***** Deliverable 5 *****/
+
+let nonRaffySighting = document.querySelector(`li[id='3']`)
```
ihollander, line 49: If you use the `data-id` instead of `id`, here's how this selector would look:

```js
let nonRaffySighting = document.querySelector(`li[data-id='3']`)

```

ihollander:
Nice work!

PR #9, opened 2020-12-11
Josh's finished Dom Manipulation Code 

jcastro506:


index.js
```diff
@@ -3,16 +3,49 @@ const header = document.querySelector("#header")
 console.log("Here's your header:", header)
 
 
+
 /***** Deliverable 2 *****/
 
+let headerColor = document.getElementById('header')
+headerColor.style.color = 'blue';
```
ihollander, line 10: Here, you can use the `header` variable from Deliverable 1 instead (but your code works too)!

index.js
```diff
@@ -3,16 +3,49 @@ const header = document.querySelector("#header")
...
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+let foundDiv = document.querySelector('.traveler')
+    foundDiv.innerHTML +=
+    `<div id="profile">
+      <img src="${traveler.photo}"
+        alt="${traveler.name}">
+      <h2>${traveler.name}</h2>
+      <em>${traveler.nickname}</em>
+    </div>`
```
ihollander, line 23: Here, you don't want to use `+=`, since there is already some content in the `.traveler` element. A better approach for this problem is to find each element in the HTML and update its properties, like this:

```js
const img = document.querySelector("#profile img");
img.src = traveler.photo;
img.alt = traveler.name;

const name = document.querySelector("#profile h2");
name.textContent = traveler.name;

const nickname = document.querySelector("#profile em");
nickname.textContent = traveler.nickname;
```

index.js
```diff
@@ -3,16 +3,49 @@ const header = document.querySelector("#header")
...
 
 /***** Deliverable 4 *****/
 
+traveler.animalSightings.forEach(function (postObj){
+    foundDiv.innerHTML += 
+`<li data-id="${postObj.id}">
+  <p>${postObj.description}</p>
+  <img src="${postObj.photo}" alt="${postObj.species}"/>
+  <a href="${postObj.link}" target="_blank">Here's a video about the ${postObj.species} species!</a>
+</li>`
```
ihollander, line 39: This is a situation where `+=` will work! However, you'd want to use the `ul#animals` as the parent element for each of the `<li>` elements, not the `div.traveler`.

One word of caution: Just be careful of using `innerHTML +=` since it _destroys_ any existing DOM elements inside the element. This can cause have unintended consequences when we start adding event listeners to DOM elements.

index.js
```diff
@@ -3,16 +3,49 @@ const header = document.querySelector("#header")
...
+
+})
+
 
 
 /***** Deliverable 5 *****/
+
+let theId = $(this).data-id = 3
+$(this).data-item-id = 3  
+$('[data-id="3"]')
```
ihollander, line 49: Here's how to use a query selector for a data- attribute:

```js
document.querySelector('[data-id="3"]')
```

This is called the [attribute selector](https://www.w3schools.com/css/css_attribute_selectors.asp) syntax.

Also, any time you do a Google search and find code with a `$` in it, that usually means the code is using **jQuery** (an old but popular JS library) instead of vanilla JS. So anything methods with a $ won't work in our code.

ihollander:
Nice work!

PR #11, opened 2020-12-11
done

boconnell2:
'

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,18 +1,58 @@
 /***** Deliverable 1 *****/
 const header = document.querySelector("#header")
-console.log("Here's your header:", header)
+// console.log("Here's your header:", header)
 
 
 /***** Deliverable 2 *****/
-
+let headerElement = document.querySelector('#header')
+headerElement.style.color = 'blue'
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
 
+// update image
+let imgElement = document.querySelector('img')
+imgElement.src = traveler.photo
+imgElement.alt = 'raffy picture'
+
+// update name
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,18 +1,58 @@
 /***** Deliverable 1 *****/
 const header = document.querySelector("#header")
-console.log("Here's your header:", header)
+// console.log("Here's your header:", header)
 
 
 /***** Deliverable 2 *****/
-
+let headerElement = document.querySelector('#header')
+headerElement.style.color = 'blue'
```
ihollander, line 8: Your code works fine, but there's no need to create a new variable here - you can just use the `header` variable from deliverable 1!

index.js
```diff
@@ -1,18 +1,58 @@
...
+nicknameElement.textContent = traveler.nickname
 
 /***** Deliverable 4 *****/
+let parent = document.querySelector('ul')
+
+traveler.animalSightings.forEach(function(animalSighting) {
+
+    // create parent li
+    let li = document.createElement('li')
+    li.setAttribute('data-id', animalSighting.id)
```
ihollander, line 34: For dataset attributes, you can also use this syntax:

```js
li.dataset.id = sighting.id
```

```html
<li data-id="1">...</li>
```

index.js
```diff
@@ -1,18 +1,58 @@
...
+    let img = document.createElement('img')
+    img.src = animalSighting.photo
+    img.alt = animalSighting.species
+    li.append(img)
+
+    // create a 
+    let a = document.createElement('a')
+    a.href = animalSighting.link
+    a.textContent = `Here's a video about the ${animalSighting.species} species!`
+    li.append(a)
```
ihollander, line 52: Using `.append` also lets you append multiple elements at once, if you prefer:

```js
li.append(p, img, a)
```

ihollander:
Nice work!

PR #12, opened 2020-12-11
Done ! 

Ekenayy:


index.html
```diff
@@ -5,8 +5,8 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
-  <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="data.js" defer></script>
```

index.js
```diff
@@ -5,14 +5,48 @@ console.log("Here's your header:", header)
...
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
-
+console.log('This is what the traveler object looks like: ', traveler)
+profilediv = document.querySelector('div#profile')
+profilediv.children[1].innerText = traveler.name
+profilediv.children[0].src = traveler.photo
+profilediv.querySelector('em').innerText = traveler.nickname
+profilediv.querySelector('img').alt = traveler.name
```
ihollander, line 17: It's best to avoid using `.children` when accessing DOM elements, since it's not as explicit. Your approach using querySelector works well - you can use that same approach for all of this:

```js
profilediv.querySelector('h2').innerText = traveler.name
profilediv.querySelector('em').innerText = traveler.nickname
profilediv.querySelector('img').src = traveler.photo
profilediv.querySelector('img').alt = traveler.name
```

index.js
```diff
@@ -5,14 +5,48 @@ console.log("Here's your header:", header)
...
+
+
+
+sightings.forEach(function (sighting) {
+    let li = document.createElement('li')
+    let p = document.createElement('p')
+    let img = document.createElement('img')
+    let a = document.createElement('a')
+
+    li.setAttribute('data-id', `${sighting.id}`)
```
ihollander, line 35: For dataset attributes, you can also use this syntax:

```js
li.dataset.id = sighting.id
```

```html
<li data-id="1">...</li>
```

ihollander:
Excellent work!

PR #13, opened 2020-12-11
Done - Jason

JasonAJordan:


index.js
```diff
@@ -2,17 +2,72 @@
...
+let animals = document.querySelector("ul#animals")
+//console.log(animals)
+let li = document.createElement('li')
+let p = document.createElement('p')
+let img = document.createElement('img')
+let a = document.createElement('a')
+
+li.setAttribute('data-id', `${sightingsObj.id}`)
+
+li.id = sightingsObj.id
```
ihollander, line 50: For dataset attributes, you can also use this syntax:

```js
li.dataset.id = sighting.id
```

```html
<li data-id="1">...</li>
```

Also, setting the `li.id` isn't necessary, since it's not part of this deliverable.

index.js
```diff
@@ -2,17 +2,72 @@
...
+
+li.id = sightingsObj.id 
+p.textContent = sightingsObj.description
+img.src = sightingsObj.photo
+img.alt = sightingsObj.species
+a.href = sightingsObj.link 
+a.target = "_blank"
+a.textContent = `Here's a video about the ${sightingsObj.species} species!`
+
+let animalSighting = document.querySelector("ul#animals")
```
ihollander, line 58: You already have a variable for this higher up in your function:

```js
let animals = document.querySelector("ul#animals")
```

index.js
```diff
@@ -2,17 +2,72 @@
...
+li.append(p, img, a)
+animalSighting.append(li)
+})
```
ihollander, line 61: Any time you write a function or a block of code, make sure it's indented! It'll help to get in the habit of properly indenting your code early, since we'll get to writing more complex, nested code later.

ihollander:
Nice work!

PR #15, opened 2020-12-11
Done(ish)

ARElton:


index.js
```diff
@@ -1,18 +1,44 @@
...
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+
+// name
+let travelerName = document.querySelector("h2")
+travelerName.textContent = `${traveler.name}`
+
+// nickname
+let travelerNickname = document.querySelector("em")
+travelerNickname.textContent = `${traveler.nickname}`
```
ihollander, line 20: String interpolation is not necessary here - you should only use string interpolation when you're creating a string our of a variable **and** some additional content, like this:

```js
travelerName.textContent = `My name is ${traveler.name}`
```

For the code above, this will work just fine without interpolation:

```js
let travelerName = document.querySelector("h2")
travelerName.textContent = traveler.name

let travelerNickname = document.querySelector("em")
travelerNickname.textContent = traveler.nickname
```

index.js
```diff
@@ -1,18 +1,44 @@
...
+
+// let travelerImageSrc = document.querySelector("img").src 
+// travelerImageSrc.textContent = `${traveler.photo}`
+
+document.querySelector("img").src = `${traveler.photo}`
+
+// The commented out version is how I feel like this should work, but it isn't working. I can only 
+// get this to work with the above line of code, which feels icky to me. Is this what it should be?
+
+document.querySelector("img").alt = `${traveler.nickname}`
```
ihollander, line 32: The reason this line doesn't work: 

```js
let travelerImageSrc = document.querySelector("img").src 
```

is because you're finding the element, and then accessing its 'src' attribute, and saving its `src` to a variable. You want to save the whole DOM element to a variable, like this:

```js
let travelerImage = document.querySelector("img")
```

In which case you can then set its `src` and `alt` attributes:

```js
travelerImage.src = traveler.photo
travelerImage.alt = traveler.nickname
```

index.js
```diff
@@ -1,18 +1,44 @@
...
 
+// Same as the above comment. This feels weird to change things directly, rather than through a variable,
+// but for some reason, the variable wasn't working. 
 
 /***** Deliverable 4 *****/
 
+traveler.animalSightings.forEach ( function (sightings) {
+    let li = document.createElement('li') 
 
+)}
```
ihollander, line 42: You've got the right setup here! Here's one solution:

```js
traveler.animalSightings.forEach(function (sighting) {
  const li = document.createElement("li");
  const img = document.createElement("img");
  const p = document.createElement("p");
  const a = document.createElement("a");

  li.dataset.id = sighting.id;
  p.textContent = sighting.description;
  img.src = sighting.photo;
  img.alt = sighting.species;
  a.href = sighting.link;
  a.textContent = `Here's a video about the ${sighting.species} species!`;

  li.append(p, img, a);

  sightingsUl.append(li);
});
```

ihollander:
Nice work getting started on this!

PR #16, opened 2020-12-11
Beksultan

Beka23:


index.js
```diff
@@ -2,17 +2,44 @@
...
+console.log('This is what the traveler object looks like: ', traveler)
+
+let pImg = document.querySelector("#profile img")
+let pName = document.querySelector("#profile h2")
+let pNickname = document.querySelector("#profile em")
+
+pImg.src = traveler.photo
+pImg.alt = "Raffy Pic"
+pName.innerHTML = traveler.name
+pNickname.innerHTML = traveler.nickname
```
ihollander, line 20: `innerHTML` works here, but in general it's better to only use `innerHTML` if you're creating DOM elements, like this:

```js
pName.innerHTML = `<span>traveler.name</span>`
```

For just inserting plain strings of text, using `textContent` is better:

```js
pName.textContent = traveler.name
pNickname.textContent = traveler.nickname
```

index.js
```diff
@@ -2,17 +2,44 @@
...
+    img.src = sight.photo
+    img.alt = sight.species
+    a.href = sight.link
+    a.textContent = `Here's a video about the ${sight.species} species!`
+
+    let listings = document.querySelector("#animals")
 
+    li.append(p, img, a)
+    listings.append(li)
+} )
```
ihollander, line 43: Nice work on this one!

ihollander:
Great work!

PR #18, opened 2020-12-11
Done

k-will22:
Didn't finish deliverable 4 or get to 5

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,13 +5,32 @@ console.log("Here's your header:", header)
 
 /***** Deliverable 2 *****/
 
-
+document.querySelector("#header").style.color = "teal"
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
 
+document.querySelector("h2").textContent = traveler.name
+document.querySelector("em").textContent = traveler.nickname 
+document.querySelector("img").src = traveler.photo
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -5,13 +5,32 @@ console.log("Here's your header:", header)
...
+let species = traveler.animalSightings.species
+let link = traveler.animalSightings.link
+
+traveler.animalSightings.forEach(function (postObj) {
+    ani.innerHTML +=
+    `<li data-id="${id}">
+  <p>${description}</p>
+  <img src="${photo}" alt="${species}"/>
+  <a href="${link}" target="_blank">Here's a video about the ${species} species!</a>
+</li>`
```
ihollander, line 32: You're on the right track here, but since `traveler.animalSightings` is an array, calling `traveler.animalSightings.id` won't work. That's why `.forEach` is helpful - it'll give you access to each element from the array:

```js
traveler.animalSightings.forEach(function (sighting) {
    ani.innerHTML +=
    `<li data-id="${sighting.id}">
  <img src="${sighting.photo}" alt="${sighting.species}"/>
  <a href="${sighting.link}" target="_blank">Here's a video about the ${sighting.species} species!</a>
</li>`
```

Also, be careful of using `innerHTML +=` since it _destroys_ any existing DOM elements inside the element. This can cause have unintended consequences when we start adding event listeners to DOM elements.

ihollander:
Nice work!

PR #21, opened 2020-12-11
done

jyliu17:


jyliu17:
Done

index.js
```diff
@@ -1,18 +1,54 @@
-/***** Deliverable 1 *****/
+        /***** Deliverable 1 *****/
 const header = document.querySelector("#header")
 console.log("Here's your header:", header)
 
 
 /***** Deliverable 2 *****/
 
-
+header.style.color = "gold"
```
ihollander, line 8: Shiny!

index.js
```diff
@@ -1,18 +1,54 @@
...
+a.href = sighting.link
+a.textContent = `Here's a video about the ${sighting.species} species!`
+    
+    
+    
+       
+let animalList = document.querySelector("#animals")
+li.append(p, img, a)
+animalList.append(li)
+})
```
ihollander, line 49: Nice work on this! Just make sure to pay attention to indentation -- it's good to get in the habit of properly indenting code now, since we'll be dealing with more complex and potentially deeply nested code later on.

index.js
```diff
@@ -1,18 +1,54 @@
...
+em.textContent = traveler.nickname
 
 /***** Deliverable 4 *****/
 
 
+const ul = document.querySelector("#animals")
+traveler.animalSightings.forEach( function (sighting) {
+    
+let li = document.createElement("li")
+li.id = sighting.id
```
ihollander, line 30: Here, your goal is to use the `data-id` attribute, not the `id` -- they will show up differently in the HTML. Here's how you can use `data-id`:

```js
li.dataset.id = sighting.id
```

This will make the following element:

```html
<li data-id="3">
```

index.js
```diff
@@ -1,18 +1,54 @@
...
+    
+    
+       
+let animalList = document.querySelector("#animals")
+li.append(p, img, a)
+animalList.append(li)
+})
 
 /***** Deliverable 5 *****/
+let wrongPic = document.getElementById('3')
```
ihollander, line 52: If you use the data-id attribute instead of id, here's what the querySelector looks like:

```js
document.querySelector("[data-id='3']")
```

ihollander:
Nice work!

PR #22, opened 2020-12-11
done

kelseamc:


index.js
```diff
@@ -5,14 +5,42 @@ console.log("Here's your header:", header)
...
-/***** Deliverable 4 *****/
+    profileDiv.id = ("profile")
+    img.src = traveler.photo
+    img.alt = traveler.name
+    name.textContent = traveler.name 
+    nickname.textContent = traveler.nickname
+    
+    let profiles = document.querySelector("#profile")
+    profileDiv.append(img, name, nickname)
+    profiles.append(profileDiv)
```
ihollander, line 26: For this deliverable, the idea is to find the *existing* DOM elements that are already part of the HTML instead of generating new DOM elements. Here's a solution that doesn't involve creating new elements:

```js
const img = document.querySelector("#profile img");
img.src = traveler.photo;
img.alt = traveler.name;

const name = document.querySelector("#profile h2");
name.textContent = traveler.name;

const nickname = document.querySelector("#profile em");
nickname.textContent = traveler.nickname;
```

index.js
```diff
@@ -5,14 +5,42 @@ console.log("Here's your header:", header)
...
+/***** Deliverable 4 *****/
+let animals = document.querySelector("#animals")
+    traveler.animalSightings.forEach(function(animal){
+        animals.insertAdjacentHTML('beforeend', `
+        
+            <li data-id= "${animal.id}">
+            <p>{animal.description}</p>
+            <img src="${animal.photo}" alt="${animal.species}"/>
+            <a href="${animal.link}" target="_blank">Here's a video about the ${animal.species} species!</a>
+        </li>  `)
```
ihollander, line 38: Great work on this! `insertAdjacentHTML` is a nice approach to creating DOM elements from strings. It's also more flexible than using `innerHTML +=`, since that destroys existing DOM elements and overwrites them. Good find!

ihollander:
Great work!

PR #26, opened 2020-12-13
done

jonathanx111:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```
BrendaMichelle, line 9: This is my preferred way of making sure the DOM is fully loaded before the Javascript code is executed. Other ways would be to move the script tag to the bottom of the body tag or adding an DOMContentLoaded event listener on the document 💯

index.js
```diff
@@ -1,18 +1,47 @@
...
+travPhoto.src = traveler.photo
+travPhoto.alt = traveler.name 
 
 /***** Deliverable 4 *****/
+list = document.querySelector("#animals")
+
+traveler.animalSightings.forEach(function(obj){
+
+    const li = document.createElement("li")
+    li.setAttribute("data-id", obj.id)
```
BrendaMichelle, line 29: For dataset attributes, you could also use this syntax:

```js
li.dataset.id = animalObj.id
```

This will add a `data-id` attribute to the element like this:

```html
<li data-id="1">...</li>
```

index.js
```diff
@@ -1,18 +1,47 @@
...
+travNickName.textContent = traveler.nickname
 
+const travPhoto = document.querySelector("#profile img")
+travPhoto.src = traveler.photo
+travPhoto.alt = traveler.name 
 
 /***** Deliverable 4 *****/
+list = document.querySelector("#animals")
+
+traveler.animalSightings.forEach(function(obj){
```
BrendaMichelle, line 26: This is good. It would be better if you had a more descriptive name for the parameter, however. Instead of `obj`, maybe `animalObj`or `sightingObj`

index.js
```diff
@@ -1,18 +1,47 @@
...
+    const p = document.createElement("p")
+    p.textContent = obj.description
 
+    const img = document.createElement("img")
+    img.src = obj.photo 
+    img.alt = obj.species 
 
+    const a = document.createElement("a")
+    a.href = obj.link 
+    a.textContent = `Here's a video about the ${obj.species} species!`
```
BrendaMichelle, line 40: Nice! Don't forget the anchor tag's target attribute. This can be set like this:

```javascript
a.target = "_blank"
```

This will tell the browser to open the link in a new tab when the user clicks on the link.

README.md
```diff
@@ -71,7 +71,7 @@ Using the `traveler` object, update the DOM to show the traveler's *name*, *nick
 
 **YOUR NOTES**
 ```
-
+I tried playing around with the order of the scripts in index.html and it seems that as long as data.js is ran before index.js, index.js will have access to the variables in data.js.
```
BrendaMichelle, line 74: Great observation!

BrendaMichelle (commented):
Nice work! Left some comments

PR #27, opened 2020-12-14
done to the best of my ability

anberjam:
Hi Michelle,

I know I'm submitting this so late but I took most of the day off on Friday since I wasn't feeling well. I worked on the mini challenge this weekend and just finished it. Would you please please mind still giving me a little bit of feedback on it whenever you get the chance? I'd appreciate it a lot. 

Anber Jamshed

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
 
   <title>Travel</title>
 </head>
@@ -28,3 +28,4 @@
 </body>
 
 </html>
+
--- a/index.js
+++ b/index.js
@@ -1,18 +1,63 @@
 /***** Deliverable 1 *****/
 const header = document.querySelector("#header")
+
+
 console.log("Here's your header:", header)
+
+//in index.html, the index.js file script was written above where the 
+//header was declared, so I added a defer in index.js script so that it gets run last
+
 
 
 /***** Deliverable 2 *****/
-
+//header is the variable assigned to that line
+header.style.color = "pink"
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+ console.log('This is what the traveler object looks like: ', traveler)
 
+ //this line finds the entire profile of the traveler 
+ travelor1 = document.querySelector("#profile")
+ //this line finds the line with the travelers name. Used h2 tag to find it
+ //cuz it's the only h2 in profile
+ travelorName = travelor1.querySelector('h2') 
+ //use textContent to change the content in between the h2 tags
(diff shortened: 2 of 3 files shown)


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```
BrendaMichelle, line 9: This is my preferred way of making sure the DOM is fully loaded before the Javascript code is executed. Other options would be to move the script tag to the bottom of the body tag or adding an DOMContentLoaded event listener 💯

index.js
```diff
@@ -1,18 +1,63 @@
...
+ travelorName = travelor1.querySelector('h2') 
+ //use textContent to change the content in between the h2 tags
+ travelorName.textContent = traveler.name
+ travelorNickname = travelor1.querySelector('em')
+ travelorNickname.textContent = traveler.nickname
+
+ travelorPic = travelor1.querySelector('img')
+ //gotta change the src and alt !! 
+ travelorPic.src = traveler.photo
+ travelorPic.alt = traveler.name
```
BrendaMichelle, line 33: Nice job scoping your search for the elements!

index.js
```diff
@@ -1,18 +1,63 @@
...
 /***** Deliverable 4 *****/
+animals = document.querySelector("#animals")
 
+traveler.animalSightings.forEach(function(postObj) {
+let li = document.createElement ('li')
+let p = document.createElement ('p')
+let img = document.createElement ('img')
+let a = document.createElement ('a')
+
+li.classList.add("data-id")
```
BrendaMichelle, line 44: The data attribute is different from a class. To add a data attribute to the li element, you can do `li.dataset.id = postObj.id`. Later on, this will be useful to help us uniquely identify which li and which animal sighting a user clicked on!

index.js
```diff
@@ -1,18 +1,63 @@
...
+li.textContent = postObj.id
+p.textContent = postObj.description
+img.src = postObj.photo
+img.alt = postObj.species
+a.setAttribute('href', `${postObj.link}`);
+a.setAttribute('target', '_blank');
```
BrendaMichelle, line 50: Nice! `href` and `target` are also defined on the `a` element, so you could also do:
```javascript
a.href = postObj.link
a.target = '_blank'
```

index.js
```diff
@@ -1,18 +1,63 @@
...
+
+li.append(p, img, a)
+animals.append(li)
+})
 
 
 /***** Deliverable 5 *****/
+//I know you have to find the element using querySelector and then use .remove()
+//I just don't know how to search for it though!!
+
```
BrendaMichelle, line 61: This is a tricky CSS selector! The key here was setting the data id attribute correctly in the previous deliverable, then searching for it using that data id!
```javascript
postToRemove.remove()
```

BrendaMichelle (commented):
Great work! Left some comments

PR #28, opened 2020-12-14
done

dashdavid13:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer ></script>
```

index.js
```diff
@@ -5,14 +5,31 @@ console.log("Here's your header:", header)
...
-
+header.style.color = "blue"
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
 
+const img = document.querySelector(".traveler img")
+const h2 = document.querySelector(".traveler h2")
+const em = document.querySelector(".traveler em")
```
BrendaMichelle, line 15: Nice, this works! An even more specific CSS selector in this case would be `"#profile img"`, `"#profile h2`", etc.

index.js
```diff
@@ -5,14 +5,31 @@ console.log("Here's your header:", header)
...
+<p>${animalSighting.description}</p>
+ <a href="${animalSighting.link}" target="_blank">Here's a video about the ${animalSighting.link} species!</a>`
+ animalsUl.append(li)
+}
 
+traveler.animalSightings.forEach(addSighting)
 
 
-/***** Deliverable 5 *****/
+/***** Deliverable 5 *****s/
```
BrendaMichelle, line 35: This one is dependent on setting the dataset attribute correctly in deliverable 4, which it looks like you did! The solution to this deliverable would look like this:

```javascript
animalToRemove.remove()
```

BrendaMichelle (commented):
Great work! Left some comments

PR #29, opened 2020-12-14
not yet complete

saracemal:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```
BrendaMichelle, line 9: Perfect! Other ways would be to move the script tag to the bottom of the body tag or adding an DOMContentLoaded event listener 💯

index.js
```diff
@@ -1,18 +1,65 @@
...
+
+// called style.color on the header and changed the color
 
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+// my guess as to why you can access traveler in the index.js is because data.js is added into the index.html inside of a script tag directly above index.js(?)
+
```
BrendaMichelle, line 24: Exactly, great guess :D

index.js
```diff
@@ -1,18 +1,65 @@
...
+/////name, nickname, photo
+
+let div = document.querySelector('div')
```
BrendaMichelle, line 27: this will get you the `<div class="traveler">...</div>` element. It already has a class of traveler, so no need to add it again on line 33

index.js
```diff
@@ -1,18 +1,65 @@
...
+let imageDiv = document.querySelector('div')
```
BrendaMichelle, line 28: This will get you the same element as above. It doesn't seem you're using this anywhere, either, so you could probably just delete this line

index.js
```diff
@@ -1,18 +1,65 @@
...
+let img = document.querySelector('img')
```
BrendaMichelle, line 29: This works! If you want a little bit more of a specific CSS selector, you could use:

```javascript
let imgTag = document.querySelector('#profile img')
```
So here, I'm searching for an `img` tag that's nested inside an element with the ID of `profile`. You could do similarly for lines 30 & 31 if you wanted a bit more of a specific CSS selector!

index.js
```diff
@@ -1,18 +1,65 @@
...
+let newH2 = document.querySelector('h2')
+let newEm = document.querySelector('em')
+// 
+div.classList.add("traveler")
```
BrendaMichelle, line 33:
```suggestion
```

index.js
```diff
@@ -1,18 +1,65 @@
...
+img.src = traveler.photo
```
BrendaMichelle, line 34: Good! Don't forget to add the `alt` attribute for the image:

```javascript
img.alt = traveler.name
```

index.js
```diff
@@ -1,18 +1,65 @@
...
+// newH2.classList.add("name")
+newH2.textContent = traveler.name
+newEm.textContent = traveler.nickname
 
 
 /***** Deliverable 4 *****/
 
+traveler.forEach(function (animalSighting) {
```
BrendaMichelle, line 42:
```suggestion
traveler.animalSightings.forEach(function (animalSighting) {
```

index.js
```diff
@@ -1,18 +1,65 @@
...
+    let li = document.createElement('li')
+    let p = document.createElement('p')
+    let img = document.createElement('img')
+    let alt = document.createElement('alt')
```
BrendaMichelle, line 46: You can remove this line; `alt` is an attribute of an `img` tag

index.js
```diff
@@ -1,18 +1,65 @@
...
+    let link = document.createElement('href')
```
BrendaMichelle, line 47: Almost! So `href` is an attribute of the `a` HTML tag. So you can write:

```javascript
let a = document.createElement('a') // creating the HTML anchor tag
```

We use the anchor tag (`<a></a>`) for adding links to our web pages! The `href` attribute on an anchor tag is the URL that we want this to be a link to!

index.js
```diff
@@ -1,18 +1,65 @@
...
+
+    li.className.add("data-id")
```
BrendaMichelle, line 49:
```suggestion
    li.dataset.id = animalSighting.id
```

dataset attributes are different than classes and ids!

index.js
```diff
@@ -1,18 +1,65 @@
...
+    p.textContent = traveler.animalSighting.description
+    img.src = traveler.animalSighting.photo
+
+
```
BrendaMichelle, line 53: Almost!

```javascript
p.textContent = animalSighting.description

img.src = animalSighting.photo
img.alt = animalSighting.species

a.target = '_blank'
a.href = animalSighting.link
a.textContent = `Here's a video about the ${animalSighting.species} species!`
```

index.js
```diff
@@ -1,18 +1,65 @@
...
+{/* <li data-id="{animalSighting id}">
+  <p>{animal sighting description}</p>
+  <img src="{animal sighting photo}" alt="{animal sighting species}"/>
+  <a href="{animal sighting link}" target="_blank">Here's a video about the {animal sighting species} species!</a>
+</li> */}
 
 
 /***** Deliverable 5 *****/
```
BrendaMichelle, line 65: This deliverable is dependent on setting the data attribute correctly in deliverable 4. To access the div with the data-id of 3 and remove it, you could write:

```javascript
animalToRemove.remove()
```

PR #30, opened 2020-12-15
first commit

BMoClay:


index.js
```diff
@@ -5,14 +5,21 @@ console.log("Here's your header:", header)
 
 /***** Deliverable 2 *****/
 
-
+header.style.color = "Blue"
```
BrendaMichelle, line 8: Good! As long as you have the `defer` attribute on the HTML script element and `const header = document.querySelector("#header")` above this, this will work. These HTML nodes have a `style` property that we can use to set some style programmatically

index.js
```diff
@@ -5,14 +5,21 @@ console.log("Here's your header:", header)
...
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
-
+console.log('This is what the traveler object looks like: ', traveler)
+let img = document.querySelector(#profile.photo)
```
BrendaMichelle, line 12: - The CSS selector that we pass as an argument to `querySelector` has to be in quotes. 

- Next - the CSS selector you wrote is looking for an element that has an id of 'profile' AND a class of 'photo', so essentially some element that looks likes this"
```HTML
<element id="profile" class="photo">
```

But first, there's no element on our page that contains a class of `photo`. I'm assuming you're interested in grabbing the `img` tag that's nested inside the profile div, so in order to follow along with where you were trying to go with this, go to the HTML file and locations this element: 
```HTML
img src="https://image.shutterstock.com/image-vector/ui-image-placeholder-wireframes-apps-260nw-1037719204.jpg" alt="{traveler name}">
```

 Inside this element add a class called `photo`. It should look something like this after the update:


```HTML
img src="https://image.shutterstock.com/image-vector/ui-image-placeholder-wireframes-apps-260nw-1037719204.jpg" alt="{traveler name}" class='photo'>
```

Ok, so we're looking for an element with a class of 'photo' that's nested somewhere inside another element that has an id of 'profile'. So, essentially, we're looking for something like this:

```HTML
<element id="profile">
  …
    <element class="photo">
     <element>
</element>
```

The CSS selector for this is super close to what you had! You need a space between `#profile` and `.photo`. This space indicates that the right hand side value (`.photo`) is nested inside the left hand side value (`#profile`)
```suggestion
let img = document.querySelector('#profile .photo')
```

index.js
```diff
@@ -5,14 +5,21 @@ console.log("Here's your header:", header)
...
+let h2 = document.querySelector(#profile.name)
+let em = document.querySelector(#profile.nickname)
```
BrendaMichelle, line 14: Same thing here. Add the `name` and `nickname` classes to the appropriate elements inside the HTML file to make the suggestion below work.

```suggestion
let h2 = document.querySelector('#profile .name')
let em = document.querySelector('#profile .nickname')
```

If you want to try to search for the `h2` and the `em` without adding a class in the CSS, you could also do:

```suggestion
let h2 = document.querySelector('#profile h2')
let em = document.querySelector('#profile em')
```

So in the first line of code in the suggestion above -- I'm searching for an `h2` type of element that's nested inside another element with the id of `profile`. Similarly for the second line -- 'm searching for an `em` type of element that's nested inside another element with the id of `profile`.

index.js
```diff
@@ -5,14 +5,21 @@ console.log("Here's your header:", header)
...
```
BrendaMichelle, line 15: So once you make the updates above, try console logging all those variables and make sure they're correlating with the appropriate element. Afterwards, take a look inside the HTML file and locate the `{ }` inside the profile div. These are the sections we need to update using DOM manipulation.

So if you have access to an element, you can change its `textContent`. You can get the text that we should update the text content with from inside the `traveler` object variable that's located inside the `data.js` file. The variable from that file is accessible to us inside the `index.js` file because it has its own script tag in the HTML file located above the script tag for `index.js`

index.js
```diff
@@ -5,14 +5,21 @@ console.log("Here's your header:", header)
...
 
 /***** Deliverable 4 *****/
-
-
+let animalSightingList = 
+<li data-id="{animalSighting id}">
+  <p>{animal sighting description}</p>
+  <img src="{animal sighting photo}" alt="{animal sighting species}"/>
+  <a href="{animal sighting link}" target="_blank">Here's a video about the {animal sighting species} species!</a>
+</li>
```
BrendaMichelle, line 23: So the `traveler` object variable contains the `animalSightings` array, which we can access using dot notation like this:
```javascript
traveler.animalSightings
```

The first step here is to iterate through this array. You can use [`forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) or [`for...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of). 

The block of code that we write inside the callback function if we're using `forEach` should be the code for creating one animal sighting post. You can use the artisanal way or the innerHTML way. If you're going for the artisanal approach, you're going to need [`document.createElement()`
](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement) to create the tags, then the next step should be to add the textContent and any attributes to the elements, then slap those elements on the DOM. 

If you're going the innerHTML route, get programatic to the `ul` element on the DOM (because that element should be the parent of all the animal sighting posts) and you can update the innerHTML for that element. Don't forget to wrap the HTML in back ticks (`), so that you can string interpolate data from the animal sighting object.

index.js
```diff
@@ -5,14 +5,21 @@ console.log("Here's your header:", header)
...
 
 /***** Deliverable 5 *****/
```
BrendaMichelle, line 24: For deliverable 5 - the key here is first setting the data-id property on the `li` element when you're creating it in step 4. Remember, to set a data-* property on an HTML element using Javascript, we need to use the `dataset` property on the element. For example:
```Javascript
someElement.dataset.id = 4
```

Then here, you're going to want to search for that element using `querySelector`. Then once you find it, you can remove it by calling `remove()` on that element.

BrendaMichelle (commented):
Take a look at my comments, take another jab at the tasks, attempt the last couple, and then re-submit!

PR #31, opened 2020-12-16
almost there

BMoClay:


BMoClay:
third commit this should be last commit for this lab

index.js
```diff
@@ -1,18 +1,40 @@
...
+console.log(h2)
 
-/***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+let em = document.querySelector("#profile .nickname")
+em.textContent = traveler.nickname
 
+let img = document.querySelector("#profile .photo")
+img.src = traveler.photo
+img.alt = traveler.name
```
rlc900, line 22: Awesome job! I think though you meant to target `travelerObj` instead of `traveler`. If you tested this in the console, your h2 and img won't have anything inside them! But your logic for this is there!

index.js
```diff
@@ -1,18 +1,40 @@
...
 
-/***** Deliverable 4 *****/
+//D4
+const allAnimalSightings = traveler.animalSightings
+let animalSightingList = document.querySelector('#animals')
```
rlc900, line 26: Nice work for this too! In this case `traveler` would also be `travelerObj` again :)

rlc900 (commented):
Very nice job with this challenge! The only point I'd say is to remember to test your code because it wouldn't have shown what you wanted it to, and this was a very little mistake and testing prevents this from happening! But as far as logic goes everything looks great and your code is very clean 👍

PR #36, opened 2021-01-12
Done

efeng732:
JSON

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,17 +2,46 @@
 const header = document.querySelector("#header")
 console.log("Here's your header:", header)
 
+// just added a defer 
 
 /***** Deliverable 2 *****/
+
+header.style.color = "blue";
 
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+ console.log('This is what the traveler object looks like: ', traveler)
+const newImg = document.querySelector(".traveler #profile img");
+const travelerName= document.querySelector("div.traveler div#profile h2 ");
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -2,17 +2,46 @@
...
+    const img = document.createElement('img');
+    img.src = animal.photo;
+    img.alt = animal.species; 
+    const a = document.createElement('a');
+    a.href = animal.link;
+    a.textContent = `Here's a video about the ${animal.species}!`;
+    li.append(p, img, a);
+    ul.append(li);
+
+})
```
rlc900, line 41: Yes!!

index.js
```diff
@@ -2,17 +2,46 @@
...
 
 
 /***** Deliverable 5 *****/
+
+const hmm= document.querySelectorAll("li");
+hmm[2].remove();
```
rlc900, line 47: SO close for this one! you just had to access the id by using `[data-id='3']` but you almost got it! Nice try 👍

index.js
```diff
@@ -2,17 +2,46 @@
 const header = document.querySelector("#header")
 console.log("Here's your header:", header)
 
+// just added a defer 
 
 /***** Deliverable 2 *****/
 
+header.style.color = "blue";
+
```
rlc900, line 11: Blue is my fav too!!

rlc900 (commented):
Very nice job! The only thing that looks like you had trouble with was accessing the `data-id` but other than that everything looked awesome 👍

PR #37, opened 2021-01-12
Done

zbayard:


index.js
```diff
@@ -1,18 +1,43 @@
...
 
 /***** Deliverable 3 *****/
 // console.log('This is what the traveler object looks like: ', traveler)
 
+document.querySelector('div#profile h2').innerText = traveler.name
+document.querySelector('div#profile em').innerText = traveler.nickname
+document.querySelector('div#profile img').src = traveler.photo
+document.querySelector('div#profile img').alt = traveler.name
+
+
```
rlc900, line 18: I like the way you chained the attributes to update! In other cases where you might be working with more than one of the same css attribute, you might want to save them to diff variables and update them accordingly, but your logic is correct for this! Awesome job!

index.js
```diff
@@ -1,18 +1,43 @@
...
+    `<li data-id="${sightingObj.id}">
+        <p>${sightingObj.description}</p>
+        <img src="${sightingObj.photo}" alt="${sightingObj.species}"/>
+        <a href="${sightingObj.link}" target="_blank">Here's a video about the ${sightingObj.species} species!</a>
+    </li>`
+
+
+
+
+})
```
rlc900, line 35: Nice work!

index.js
```diff
@@ -1,18 +1,43 @@
...
+
+
 
 
 /***** Deliverable 5 *****/
+
+
+document.querySelector("li[data-id='3']").remove()
```
rlc900, line 43: 👍

rlc900 (commented):
Overall awesome job! It seems like you have a pretty good understanding of the material so far! Nice work 👍

PR #41, opened 2021-01-12
Done

clairedeboer:


index.js
```diff
@@ -4,15 +4,36 @@ console.log("Here's your header:", header)
 
 
 /***** Deliverable 2 *****/
-
+header.style.color = "green"
```
BrendaMichelle, line 7: Ooo, green goes well with this theme!!

README.md
```diff
@@ -70,7 +70,7 @@ First, uncomment the `console.log` under Deliverable 3 in the `index.js` file to
 Using the `traveler` object, update the DOM to show the traveler's *name*, *nickname*, and *photo* in the appropriate places. For the `<img>` tag, make sure to update the `src` attribute *and* the `alt` attribute (it's important for accessibility).
 
 **YOUR NOTES**
-```
+```You can access traveler object from data.js file in index.js file becaues index.js inherits from data.js?
```
BrendaMichelle, line 73: We can access the traveler object from the data.js file because it's a global variable and its content is loaded in from the same HTML file as our index.js file (so they both have access to the same global scope)!

BrendaMichelle (commented):
Nice work!

PR #45, opened 2021-01-12
Stopped to pay attention to lecture

jaimedavid01:


index.html
```diff
@@ -5,8 +5,8 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
-  <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="data.js" defer></script>
+  <script src="index.js" defer></script>
```
BrendaMichelle, line 10: Good! The script tag for the data.js file doesn't technically need a `defer` because it doesn't contain any code that will try to access elements from the DOM, but it's okay either way in this case!

index.js
```diff
@@ -5,14 +5,50 @@ console.log("Here's your header:", header)
...
+
+    ul.append(newLi)
+
+
+})
+
+
 
 
 /***** Deliverable 5 *****/
```
BrendaMichelle, line 54: You can do it (:

index.js
```diff
@@ -5,14 +5,50 @@ console.log("Here's your header:", header)
...
+    newImg.alt = animal.species
+    
+    let newLink = document.createElement('a')
+    newLink.href = animal.newLink
+    newLink.target = "_blank"
+    newLink.textContent = `Here's a video about the ${animal.species} species!`
+
+    newLi.append(newP, newImg, newLink)
+
+    ul.append(newLi)
```
BrendaMichelle, line 46: Looks good. Semantic variable naming, organized well 💯

index.js
```diff
@@ -5,14 +5,50 @@ console.log("Here's your header:", header)
...
+
+photo.src = traveler.photo
+newName.textContent = traveler.name
+newNickName.textContent = traveler.nickname
+
 
 
 /***** Deliverable 4 *****/
 
+const ul = document.querySelector('ul#animals')
```
BrendaMichelle, line 26: Smart idea writing this outside of the callback function for `forEach`. Saves computational time & energy!

PR #48, opened 2021-01-12
all deliverables finishedd

jpersons24:
another one!

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,7 +6,6 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
 
   <title>Travel</title>
 </head>
@@ -25,6 +24,8 @@
       <!-- add an <li> for each animal sighting here -->
     </ul>
   </div>
+
+  <script src="index.js"></script>
 </body>
 
 </html>
--- a/index.js
+++ b/index.js
@@ -4,15 +4,72 @@ console.log("Here's your header:", header)
 
 
 /***** Deliverable 2 *****/
-
+header.style.color = 'blue' 
 
 
 /***** Deliverable 3 *****/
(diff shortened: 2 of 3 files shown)


index.html
```diff
@@ -25,6 +24,8 @@ <h4>Great Sightings</h4>
       <!-- add an <li> for each animal sighting here -->
     </ul>
   </div>
+
+  <script src="index.js"></script>
```
Dwire, line 28: Nice placement at the bottom of the HTML page. You could have opted to keep the script tag in place and use "defer" both are fine nice job.

index.js
```diff
@@ -4,15 +4,72 @@ console.log("Here's your header:", header)
...
+// next time --> select existing elements and replace content
+profile.querySelector('img').remove()
+profile.querySelector('h2').remove()
+profile.querySelector('em').remove()
+
+profile.innerHTML +=
+   `<img src=${traveler.photo}
+   alt=${traveler.name}>
+   <h2>${traveler.name}</h2>
+   <em>${traveler.nickname}</em>`
```
Dwire, line 34: Nice work. You wrote a bit more code than needed here though. If you are going to use the .innerHTML way of solving this problem then you can just write = in this case instead of +=.  Just using = will replace all the HTML for #profile += is adding additional HTML.

get rid of lines 26 through 28 and change += to =.

index.js
```diff
@@ -4,15 +4,72 @@ console.log("Here's your header:", header)
...
+   newLi.append(newP, newImg, newLink)
+
+   ulAnimals.append(newLi)
+})
+
 
 
 /***** Deliverable 5 *****/
+
+document.getElementById('3').remove()
```
Dwire, line 75: This looks like it would work but does avoid using the dataset attribute that the directions talk about. At some point mess around with dataset it can be useful.

Dwire (commented):
Nice work. Day 2 in the books!

PR #49, opened 2021-01-12
Done with some help

DanielYankiver:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer ></script>
 
   <title>Travel</title>
 </head>
```
Dwire, line 12: Nice job catching why the two files were not on speaking terms and fixing it with defer.

index.js
```diff
@@ -5,14 +5,50 @@ console.log("Here's your header:", header)
...
 
+const trvlrH2 = document.querySelector(`h2`)
+trvlrH2.innerText = traveler.name
+
+const trvlrEm = document.querySelector(`em`)
+trvlrEm.innerText = traveler.nickname
+
+const trvlrImg = document.querySelector(`img`)
+trvlrImg.src = traveler.photo
+trvlrImg.alt = "raffy photo"
```
Dwire, line 24: We worked on this a bit together. Looks good 👌

index.js
```diff
@@ -5,14 +5,50 @@ console.log("Here's your header:", header)
...
+
+    const newA = document.createElement(`a`)
+    newA.href = sightings.link
+    newA.textContent = `Here's a video about the ${sightings.species} species!` 
+
+
+    newLi.append(newP, newImg, newA)
+    ul.append(newLi)
+})
+
```
Dwire, line 49: This looks really good nice work. What would it look like if you swapped out the callback in your forEach for an arrow function?

index.js
```diff
@@ -5,14 +5,50 @@ console.log("Here's your header:", header)
...
 
 
 /***** Deliverable 5 *****/
+
+document.querySelector(`[data-id='3']`).remove()
```
Dwire, line 54: 🔥 Already talked about datasets a little with you when we were on zoom.

PR #51, opened 2021-02-02
Josh Frank phase 3 DOM manipulation mini-challenge

josh-frank:
Completed 2/2/2021

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,13 +6,13 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
 
   <title>Travel</title>
 </head>
 
 <body>
-  <h1 id="header"> Amazon Trip</h1>
+  <h1 id="header">Amazon Trip</h1>
   <div class="traveler">
     <div id="profile">
       <img src="https://image.shutterstock.com/image-vector/ui-image-placeholder-wireframes-apps-260nw-1037719204.jpg"
--- a/index.js
+++ b/index.js
@@ -1,18 +1,40 @@
 /***** Deliverable 1 *****/
-const header = document.querySelector("#header")
-console.log("Here's your header:", header)
-
+const header = document.querySelector( "h1#header" );
+console.log( "Here's your header:", header );
 
 /***** Deliverable 2 *****/
-
-
+header.style = "color:orange;";
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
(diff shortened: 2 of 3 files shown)


index.html
```diff
@@ -6,13 +6,13 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```

index.js
```diff
@@ -1,18 +1,40 @@
...
+console.log( 'This is what the traveler object looks like: ', traveler );
+const travelerProfile = document.querySelector( "div#profile" );
+const travelerPhoto = travelerProfile.querySelector("img");
+const travelerName = travelerProfile.querySelector("h2");
+const travelerNickname = travelerProfile.querySelector("em");
 
+travelerPhoto.src = traveler.photo
+travelerPhoto.alt = `A photo of ${traveler.name}`
+travelerName.textContent = traveler.name
+travelerNickname.textContent = traveler.nickname
```
BrendaMichelle, line 18: Good work scoping your search by finding the `travelerProfile` first. Scoping our search is also better performance wise because there's less elements to traverse through

index.js
```diff
@@ -1,18 +1,40 @@
...
+    let thisSightingImage = document.createElement( "img" );
+    thisSightingImage.src = sighting.photo;
+    thisSightingImage.alt = sighting.species;
+    let thisSightingLink = document.createElement( "a" );
+    thisSightingLink.href = sighting.link;
+    thisSightingLink.target = "_blank";
+    thisSightingLink.textContent = `Here's a video about the ${ sighting.species } species!`
+    thisSighting.append( thisSightingDescription, thisSightingImage, thisSightingLink );
+    travelerSightings.appendChild( thisSighting );
+} );
```
BrendaMichelle, line 36: Great work "artisanally" creating the elements and great use of the arrow function. You can also remove the parenthesis around the parameter of the arrow function since it's only accepting exactly one argument.

index.js
```diff
@@ -1,18 +1,40 @@
...
 
 /***** Deliverable 5 *****/
+const sightingToRemove = document.querySelector( "li[data-id='3']" )
+sightingToRemove.remove();
```
BrendaMichelle, line 40: 💯

BrendaMichelle (commented):
Great work! I appreciate your use of `const`. There two trains of thought on using `const`. One is using `const` when you know for sure that a value should be constant up front, such as when you're storing some constant math variable or file version number, for example. The other train of thought is to always use `const` to declare variables and then change the declaration to `let` down the road if you realize you need to be able to re-assign a variable. It seems like you gravitate towards the second approach      💯

PR #52, opened 2021-02-02
Completed Mini-Challenge

LaurenGifford:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script defer src="index.js"></script>
```

index.js
```diff
@@ -4,15 +4,42 @@ console.log("Here's your header:", header)
...
+const profile = document.querySelector('#profile')
+const nameData = profile.querySelector('h2')
+nameData.textContent = traveler.name
+
+const nickname = profile.querySelector('em')
+nickname.textContent = traveler.nickname
 
+const image = profile.querySelector('img')
+image.src = traveler.photo
+image.alt = traveler.name
```
BrendaMichelle, line 22: Nice work scoping the search to the `profile`. Scoping our searches is also better performance wise because there's less DOM traversal needed to be done!

index.js
```diff
@@ -4,15 +4,42 @@ console.log("Here's your header:", header)
...
+    
+    sightData.innerHTML = `<li data-id=${sighting.id}>
+    <p>${sighting.description}</p>
+    <img src=${sighting.photo} alt=${sighting.species}/>
+    <a href=${sighting.link} target="_blank">Here's a video about the ${sighting.species} species!</a>
+  </li>`
+
+    const animals = document.querySelector('#animals')
+    animals.append(sightData)
+})
```
BrendaMichelle, line 39: well done!

index.js
```diff
@@ -4,15 +4,42 @@ console.log("Here's your header:", header)
...
 
 
 /***** Deliverable 5 *****/
+
+const removeEl = document.querySelector("[data-id='3']")
+removeEl.remove()
```
BrendaMichelle, line 45: 💯

BrendaMichelle (commented):
Great work! There's two trains of thought on using `const`. One is using `const` when you know for a fact that a value should be constant up front, such as when you're storing some constant math variable or file version number, for example. The other train of thought is to always use `const` to declare variables and then change the declaration to `let` down the road if/when you realize that you need to be able to re-assign a variable. It seems like you gravitate towards the second approach and I appreciate your approach  💯

PR #53, opened 2021-02-02
done

shmuly11:


index.js
```diff
@@ -5,14 +5,42 @@ console.log("Here's your header:", header)
...
+let profile = document.querySelector('#profile')
+console.log(profile)
+let name = profile.querySelector('h2')
+let nickname = profile.querySelector('em')
+let image = profile.querySelector('img')
+console.log(name, nickname, image)
+name.innerText = traveler.name
+nickname.innerText = traveler.nickname
+image.src = traveler.photo
+image.alt = traveler.name
```
BrendaMichelle, line 22: Nice work scoping your search to the `profile`! Scoping our searches is also better performance wise because there's less DOM traversal needed to be done!

index.js
```diff
@@ -5,14 +5,42 @@ console.log("Here's your header:", header)
...
+    list.dataset.id = sighting.id
+    list.innerHTML = `<p>${sighting.description}</p>
+    <img src="${sighting.photo}" alt="${sighting.species}"/>
+    <a href="${sighting.link}" target="_blank">Here's a video about the ${sighting.species} species!</a>`
+
+    console.log(list)
+
+    let animals = document.querySelector('#animals')
+    animals.append(list)
+})
```
BrendaMichelle, line 37: This is well done! One possible way to improve this performance wise would be to move line 35 to the outer scope, so that we only have to traverse the DOM once in order to find this element with the `animals` id while still having access to that variable in this callback function.

BrendaMichelle (commented):
Well done, keep up the good work!

PR #54, opened 2021-02-02
Finished

mcardona9015:


index.js
```diff
@@ -1,18 +1,49 @@
 /***** Deliverable 1 *****/
 const header = document.querySelector("#header")
+console.log(header)
 console.log("Here's your header:", header)
 
 
 /***** Deliverable 2 *****/
 
-
+header.style.color = "red"
```
BrendaMichelle, line 9: My favorite color, too 🤘

index.js
```diff
@@ -1,18 +1,49 @@
...
+
+const travelerImg = travelerDiv.querySelector("img")
+travelerImg.src = traveler.photo
+travelerImg.alt = traveler.name
 
+const travelerName = travelerDiv.querySelector("h2")
+travelerName.textContent = traveler.name
 
+const travelerNickame = travelerDiv.querySelector("em")
+travelerNickame.textContent = traveler.nickname
```
BrendaMichelle, line 23: Nice work scoping the search to the `travelerDiv`. Scoping our searches is also better performance wise because there's less DOM traversal needed to be done!

index.js
```diff
@@ -1,18 +1,49 @@
...
+    animalImg.alt = animal.species
 
+    const animalA = document.createElement("a")
+    animalA.href = animal.link
+    animalA.target = "_blank"
+    animalA.innerText = `Here's a video about the ${animal.species} species!`
 
+    animalLi.append(animalP, animalImg, animalA)
+    animals.append(animalLi)
+})
```
BrendaMichelle, line 45: Great work. Nice thinking of declaring the `animals` variable in the scope outside the callback. It improves performance only having to find this element once instead of having it inside the forEach and re-finding it over and over.

index.js
```diff
@@ -1,18 +1,49 @@
...
 
 /***** Deliverable 5 *****/
+const centipede = animals.querySelector(`[data-id='3']`)
+centipede.remove()
```
BrendaMichelle, line 49: 💯

BrendaMichelle (commented):
Great work! Clean code & well named variables. There's two trains of thought on using `const`. One is using `const` when you know for a fact that a value should be constant up front, such as when you're storing some constant math variable or file version number, for example. The other train of thought is to always use `const` to declare variables and then change the declaration to `let` down the road if/when you realize that you need to be able to re-assign a variable. It seems like you gravitate towards the second approach and I appreciate your approach 💯

PR #60, opened 2021-02-03
day2

dabryant11:
struggled on #5 - would like some further clarification if possible

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,9 +5,6 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
-  <script src="data.js"></script>
-  <script src="index.js"></script>
-
   <title>Travel</title>
 </head>
 
@@ -28,3 +25,5 @@
 </body>
 
 </html>
+<script src="data.js"></script>
+<script src="index.js"></script>
--- a/index.js
+++ b/index.js
@@ -4,15 +4,52 @@ console.log("Here's your header:", header)
 
 
 /***** Deliverable 2 *****/
-
+header.style.color = "green"
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+let headerTwo = document.querySelector("h2")
+headerTwo.innertext = traveler.name  
+
+let em = document.querySelector("em") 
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -27,4 +24,6 @@ <h4>Great Sightings</h4>
   </div>
 </body>
 
-</html>
\ No newline at end of file
+</html>
+<script src="data.js"></script>
+<script src="index.js"></script>
```
hasibulc, line 29: Good job moving the script tags down. It is usually convention to either put script tags just before the closing </body> tag or giving them an attribute of 'defer'.

index.js
```diff
@@ -4,15 +4,52 @@ console.log("Here's your header:", header)
...
-// console.log('This is what the traveler object looks like: ', traveler)
+let headerTwo = document.querySelector("h2")
+headerTwo.innertext = traveler.name  
+
+let em = document.querySelector("em") 
+em.innerText = traveler.nickname  
+
+image = document.querySelector("img")
+image.src = "pics/raffy.jpg" 
+image.alt = "raffy"
```
hasibulc, line 19: You should be using the data from 'traveler', not hard coding the image src and alt

index.js
```diff
@@ -4,15 +4,52 @@ console.log("Here's your header:", header)
...
+
+console.log('This is what the traveler object looks like: ', traveler)
 
 
 /***** Deliverable 4 *****/
 
 
+animals = document.querySelector("#animals") 
+
+traveler.animalSightings.forEach (function(animal){
```
hasibulc, line 29: good job iterating through 'taveler.animalSightings' and creating li nodes with the proper children nodes

index.js
```diff
@@ -4,15 +4,52 @@ console.log("Here's your header:", header)
...
+    li.append(img)
+    animals.append(li)
+    })
+
+
+
 
 /***** Deliverable 5 *****/
+
+animals.removeAttribute(datatset.id==3)
```
hasibulc, line 55: 'animals.attributes' returns: "NamedNodeMap {0: id, id: id, length: 1}", there is nothing that matches 'dataset.id'.
You need to select the li element and then remove it.

Possible quick solution:

const removeElem = animals.querySelector(`[data-id='3']`)
removeElem.remove()

PR #61, opened 2021-02-03
day 2 mini challenge-- very late

samhall330:


README.md
```diff
@@ -47,7 +47,7 @@ Figure out what you need to change to give Javascript access to the `h1#header`
 
 **YOUR NOTES**
 ```
-
+Adding "defer" after <script src="index.js" tells html to load everything else in the browser before running index.js. I think..?
```
hasibulc, line 50: yes!

README.md
```diff
@@ -71,6 +71,18 @@ Using the `traveler` object, update the DOM to show the traveler's *name*, *nick
 
 **YOUR NOTES**
 ```
+I'm not sure why we have access to the traveler variable declared in data.js, but I think it might be because index.html calls both js files in script tags.
+
```
hasibulc, line 75: Correct, however remember that HTML is loaded top down so the order of the script tags matter. Switching the script tag order will make it so that 'index.js' does not have access to 'data.js' since 'index.js' will be loaded before 'data.js':

 <script src="index.js"></script>

README.md
```diff
@@ -71,6 +71,18 @@ Using the `traveler` object, update the DOM to show the traveler's *name*, *nick
...
+*** These define which tags in the html doc that I'm manipulating and set them to variables ***
+const travelerPhoto = profile.querySelector("img")
+const travelerName = profile.querySelector("h2")
+const travelerNickname = profile.querySelector("em")
```
hasibulc, line 79: great variable names!

README.md
```diff
@@ -71,6 +71,18 @@ Using the `traveler` object, update the DOM to show the traveler's *name*, *nick
...
+
+*** These pull attributes from the traveler variable set in data.js and apply them to the variables I created above ***
+travelerPhoto.src = traveler.photo
+travelerPhoto.alt = "Traveler Photo"
+travelerNickname.textContent = traveler.nickname
+travelerName.textContent = traveler.name
```
hasibulc, line 85: good job accessing data using dot notation

README.md
```diff
@@ -89,7 +101,28 @@ We also want to show some of the awesome animal sightings our traveler had. You'
 
 **YOUR NOTES**
 ```
-
+traveler.animalSightings.forEach(animal =>  {
+    const newSighting = document.createElement("li")
```
hasibulc, line 105: It would be better to have the variable name reference the element you created, for ex: newSightingLi

README.md
```diff
@@ -89,7 +101,28 @@ We also want to show some of the awesome animal sightings our traveler had. You'
...
+    newSighting.dataset.id = animal.id
```
hasibulc, line 106: great job added "data-id= ' attribute

README.md
```diff
@@ -89,7 +101,28 @@ We also want to show some of the awesome animal sightings our traveler had. You'
...
+    const sightingImage = document.createElement("img")
+    sightingImage.src = animal.photo
+*** Creating <img> tag in html, setting variable for the "photo" attribute from our animalSightings array of objects. ***
+
+    const sightingLink = document.createElement("a")
+    sightingLink.href = animal.link 
+    sightingLink.textContent = `Here's a video about the ${animal.species} species!`
+*** Creating <a> tag in html, setting variable for the "link" attribute from our animalSightings array of objects, and creating an href link to another url. ***
+
+    newSighting.append(sightingDescription, sightingImage, sightingLink)
```
hasibulc, line 122: good use of append() to append multiple child nodes

PR #62, opened 2021-02-03
Finished DOM mini challenge

dsasse07:


index.html
```diff
@@ -5,8 +5,8 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
-  <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script defer src="data.js"></script>
+  <script defer src="index.js"></script>
```
hasibulc, line 9: great job adding the defer attribute to the script tags

index.js
```diff
@@ -1,18 +1,95 @@
 /***** Deliverable 1 *****/
 const header = document.querySelector("#header")
 console.log("Here's your header:", header)
-
+// Added defer to the HTML Header
 
 /***** Deliverable 2 *****/
-
+header.style.color = "rgb(0, 72, 255)"
+header.style.fontSize = "5rem"
```
hasibulc, line 8: good inline styling

index.js
```diff
@@ -1,18 +1,95 @@
...
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+
+const name = document.querySelector("#profile h2")
+name.textContent = `${traveler.name}`
+
+const nickname = document.querySelector("#profile em")
+nickname.textContent = `${traveler.nickname}`
```
hasibulc, line 18: great job accessing the traveler data. you don't need string interpolation here since the data is already a string.

index.js
```diff
@@ -1,18 +1,95 @@
...
+// Can we access data.js because it is loaded in the HTML header before index.js?
 
 /***** Deliverable 4 *****/
+console.log(traveler.animalSightings)
+const animals = document.querySelector("#animals")
+// const ul = document.createElement("ul")
 
+traveler.animalSightings.forEach(sighting => {
+  let li = document.createElement("li")
+  li.setAttribute('data-id', `${sighting.id}`)
```
hasibulc, line 38: you can also add 'data-id' attribute with '.dataset':

 li.dataset.id = sighting.id

index.js
```diff
@@ -1,18 +1,95 @@
...
+  
+  let p = document.createElement("p")
+  p.textContent = `${sighting.description}`
+  
+  let image = document.createElement("img")
+  image.setAttribute('src',`${sighting.photo}`)
+  image.setAttribute('alt', `${sighting.species}`)
```
hasibulc, line 45: You can also add 'src' and 'alt' attributes this way:
image.src = sighting.photo
image.alt = sighting.species

You should use '.setAttribute' for non standard HTML attributes. 

No need for string interpolation since 'sighting.photo' and 'sighting.species' are strings.

hasibulc:
Great code and good use of advanced solutions. However, sometimes it is better to use simpler approaches.

PR #65, opened 2021-02-23
Max-Beneke-Done

MaxBeneke:


README.md
```diff
@@ -71,7 +71,7 @@ Using the `traveler` object, update the DOM to show the traveler's *name*, *nick
 
 **YOUR NOTES**
 ```
-
+I got tripped up by a typo for way too long until I realized the 'Console' is the place that shows errors on the Chrome Dev Tools, not the 'Elements' tab. For this, I know the 'data.js' file is being loaded into the html along with the index.js and since we're manipulating the DOM (not HTML) that's why it can access the other file. But that doesn't sound too right.
```
BrendaMichelle, line 74: I'm glad you found the console tab; that must've been frustrating up to that point! Yes, the reason we can access the variable from data.js in the index.js is because the variable in data.js is a global variable that shares the same global scope of index.js because both files are loaded by the same HTML file.

README.md
```diff
@@ -89,7 +89,7 @@ We also want to show some of the awesome animal sightings our traveler had. You'
 
 **YOUR NOTES**
 ```
-
+This was tough because the forEach is still a little fuzzy in my head. I got it so that it would only print the last one, and realized setting "animal.innerHTML = <etc>" was overriding each time it went through the loop. So I found online that you can do += instead to make sure it's adding each <li> element sequentially to the innerHTML. Hope that was a good solution.
```
BrendaMichelle, line 92: Great observation and good find!!

index.html
```diff
@@ -6,20 +6,20 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```

index.js
```diff
@@ -4,15 +4,45 @@ console.log("Here's your header:", header)
...
+
+div.innerHTML = `<div id="profile">
+                <img src="${traveler.photo}"
+                alt="${traveler.name}">
+                <h2>${traveler.name}</h2>
+                <em>${traveler.nickname}</em>
+                </div>`;
+
+const travelerDiv = document.querySelector("div.traveler");
+travelerDiv.prepend(div);
```
BrendaMichelle, line 22: For this deliverable, locate the elements you want to find on the page using `querySelector` for example. Then update them. If the elements already exist on the page, it's better to just find them then update. For example:


```js
const profileImg = document.querySelector("#profile img") // find existing element on page
profileImg.src = traveler.photo // update the src attribute
profileImg.alt = traveler.name // update the alt attribute

const profileH2 = document.querySelector("#profile h2") // find existing element on page
profileH2.textContent = traveler.name // update the text displayed on the page

const profileEm = document.querySelector("#profile em") // find existing element on page
profileEm.textContent = traveler.nickname  // update the text displayed on the page

const likes = document.querySelector("#profile .likes") // find existing element on page
likes.textContent = `${traveler.likes} Likes`  // update the text displayed on the page
```

index.js
```diff
@@ -4,15 +4,45 @@ console.log("Here's your header:", header)
...
+    animal.innerHTML += `<li data-id="${sighting.id}">
+    <p>${sighting.description}</p>
+    <img src="${sighting.photo}" alt="${sighting.species}"/>
+    <a href="${sighting.link}" target="_blank">Here's a video about the ${sighting.species} species!</a>
+    </li>`
+});
+
+const animalsUl = document.querySelector('ul#animals');
+
+animalsUl.append(animal);
```
BrendaMichelle, line 41: You're on the right track! Currently you're creating an `li` and then adding a bunch of other `li`s within the `li`, so the parent of all the animal `li`s is another `li`.  You need to create [an `li`](https://www.w3schools.com/tags/tag_li.asp) for _each_ animal sighting post and the parent should be the `animalsUl`. You have several options: 

- Change the `innerHTML` of the `animalsUl`
```js
const animalsUl = document.querySelector('ul#animals');
animalsUl.append(fragment);

traveler.animalSightings.forEach( function(sighting) {
    animalsUl.innerHTML += `<li data-id="${sighting.id}">
    <img src="${sighting.photo}" alt="${sighting.species}"/>
    <a href="${sighting.link}" target="_blank">Here's a video about the ${sighting.species} species!</a>
    </li>`
});
```

- Using [document fragment](https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment) (more advanced)
```js
const fragment = document.createDocumentFragment();;

traveler.animalSightings.forEach( function(sighting) {
    fragment.innerHTML += `<li data-id="${sighting.id}">
    <img src="${sighting.photo}" alt="${sighting.species}"/>
    <a href="${sighting.link}" target="_blank">Here's a video about the ${sighting.species} species!</a>
    </li>`
});

const animalsUl = document.querySelector('ul#animals');
animalsUl.append(fragment);
```

index.js
```diff
@@ -4,15 +4,45 @@ console.log("Here's your header:", header)
...
 
 
 
 /***** Deliverable 5 *****/
+
+const intruder = document.querySelector('[data-id="3"]');
+console.log(intruder)
+intruder.remove();
```
BrendaMichelle, line 48: Good!

BrendaMichelle (commented):
Good! Left some comments. Let me know if you have any questions

PR #66, opened 2021-02-23
Completed Deliverables

gabriel-demes:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```
emotionalproof, line 9: Nice!

index.js
```diff
@@ -4,15 +4,46 @@ console.log("Here's your header:", header)
...
 /***** Deliverable 2 *****/
+header.style.color = "red"
 
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+const profile = document.querySelector("#profile")
+const name = profile.children[1];
```
emotionalproof, line 14: This works. You could also do 
travelerCard.querySelector('h2') 
this would make it more dynamic and allow you to change the order of the profile attributes without needing to change how you declare the variable

index.js
```diff
@@ -4,15 +4,46 @@ console.log("Here's your header:", header)
...
+name.innerHTML = traveler.name;
```
emotionalproof, line 15: Since you are not putting any html inside of the nickName tag, just text, you can use  .innerText

index.js
```diff
@@ -4,15 +4,46 @@ console.log("Here's your header:", header)
...
+const nickName = profile.children[2];
+nickName.innerHTML = traveler.nickname;
+const pfp = profile.children[0];
+pfp.src= traveler.photo
+pfp.alt= traveler.name
```
emotionalproof, line 20: perfect!

index.js
```diff
@@ -4,15 +4,46 @@ console.log("Here's your header:", header)
...
 
 
 /***** Deliverable 4 *****/
+const animals = document.querySelector("#animals")
+const animalSightings = traveler.animalSightings
+for(let i = 0; i <traveler.animalSightings.length; i++){
```
emotionalproof, line 26: Great! Since you already created the variable animalSightings, you can do 
i < animalSightings.length

Try refactoring this using a forEach. This would allow you to avoid referencing the index when setting attributes

index.js
```diff
@@ -4,15 +4,46 @@ console.log("Here's your header:", header)
...
+    const item = document.createElement('li');
+    item.setAttribute("data-id", `${animalSightings[i].id}`)
+    const descrip = document.createElement(`p`)
+    const descripText = document.createTextNode(`${animalSightings[i].description}`);
+    descrip.append(descripText)
```
emotionalproof, line 31: Nice use of createTextNode
It looks like you may have just done this to test out createTextNode.If that is not that case then you could just do descrip.innerText = animalSightings[i].description

index.js
```diff
@@ -4,15 +4,46 @@ console.log("Here's your header:", header)
...
+pfp.src= traveler.photo
+pfp.alt= traveler.name
 
 
 /***** Deliverable 4 *****/
+const animals = document.querySelector("#animals")
+const animalSightings = traveler.animalSightings
+for(let i = 0; i <traveler.animalSightings.length; i++){
+    const item = document.createElement('li');
+    item.setAttribute("data-id", `${animalSightings[i].id}`)
```
emotionalproof, line 28: Great! You could also do
item.dataset.id = animalSightings[i].id

index.js
```diff
@@ -4,15 +4,46 @@ console.log("Here's your header:", header)
...
+    img.src = animalSightings[i].photo;
+    img.alt = animalSightings[i].species;
+    const video = document.createElement('a')
+    const videoText = document.createTextNode(`Here's a video about the ${animalSightings[i].species} species!`)
+    video.append(videoText)
+    video.href = animalSightings[i].link
+    item.appendChild(descrip)
+    item.appendChild(img)
+    item.appendChild(video)
+    animals.appendChild(item)
```
emotionalproof, line 42: This is great!  another option would be
item.innerHTML = `
        ${sighting.species}
        <a href="${sighting.link}" target="_blank">Here's a video about the ${sighting.species} species!</a>`

emotionalproof, line 42: Nice! Try refactoring using .append which allows you to append multiple child nodes at once separated by commas

index.js
```diff
@@ -4,15 +4,46 @@ console.log("Here's your header:", header)
...
+}
 
 
 
 /***** Deliverable 5 *****/
+const unwanted = document.querySelector("[data-id='3']")
```
emotionalproof, line 48: Awesome!  Memorize this formatting. It is one of those things that is a headache to have to lookup

emotionalproof (commented):
nice!

PR #67, opened 2021-02-23
Deliverables Complete

ajperlis07:


index.js
```diff
@@ -4,15 +4,42 @@ console.log("Here's your header:", header)
...
+    var pText = document.createTextNode(value.description)
+    p.append(pText)
+    var img = document.createElement("img");
+    img.src = value.photo
+    img.alt = value.species
+    a.href = value.link
+    a.target = "_blank" 
+    var aText = document.createTextNode("Here is a video about the " + value.species + "species!")
+    a.append(aText)
+    li.setAttribute('data-id' , value.id);
```
BrendaMichelle, line 35: Good work with this deliverable! For setting dataset attributes, you can also write:
```suggestion
    li.dataset.id = value.id
```

index.js
```diff
@@ -4,15 +4,42 @@ console.log("Here's your header:", header)
...
+    li.appendChild(p);
+    li.appendChild(img);
+    li.appendChild(a)
+    document.querySelector("#animals").appendChild(li);
+    
+})
 
 
 /***** Deliverable 5 *****/
+document.querySelector("[data-id = '3']").remove()
```
BrendaMichelle, line 45: 💯

BrendaMichelle (commented):
Good work, looks good! Try to stick with using  `let` and `const` for variable declarations. It's considered best practice since ES6.

PR #72, opened 2021-02-23
William Bugenis - Done

wbugenis:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```
emotionalproof, line 9: Great!

index.js
```diff
@@ -1,18 +1,49 @@
...
+header.style.color = "aqua"
+//I used console.dir(header) in the dev console on Chrome to check for the applicable key value pairs, found the above style and color keys, and then set the color to aqua.
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+
+const travelerDiv = document.querySelector("div.traveler")
+console.log(travelerDiv)
```
emotionalproof, line 15: Love this. Always console.log before moving on to the rest of the code

index.js
```diff
@@ -1,18 +1,49 @@
...
+travelerDiv.querySelector("h2").innerText = traveler.name
+travelerDiv.querySelector("em").innerText = traveler.nickname
+travelerDiv.querySelector("img").src = traveler.photo
```
emotionalproof, line 18: Nice job! Since you are not referencing those particular tags again, there is no need to set them to variables

index.js
```diff
@@ -1,18 +1,49 @@
...
+
+//After checking out the contents of the traveler object, I found the elements on the page that needed to be changed and assigned the appropriate values.
+//Is this the most efficient way of accessing the inner elements, or is there an easier/better way to do it that I'm forgetting?
 
 
 /***** Deliverable 4 *****/
+const animalList = travelerDiv.querySelector("ul#animals")
```
emotionalproof, line 25: Perfect

emotionalproof, line 25: Perfect! No need to set this variable inside the functoin

index.js
```diff
@@ -1,18 +1,49 @@
...
+    animalList.append(li)
+    }
+)
+//After assigning the unordered animals list to a variable, I iterated through each of the animal sightings to create a list element for each, filling out the changing fields with string interpolation - then I appended the newly created list item onto the unordered list on the page.
 
 
 /***** Deliverable 5 *****/
+
+// animalList.querySelector('[data-id="3"]').remove()
+animalList.children[2].remove()
```
emotionalproof, line 43: document.querySelector('[data-id="3"]').remove() would be preferred since it would be more dynamic. Let's say you had sort or filter functionality that changed the order of the children, animalList.children[2].remove() would no longer remove the correct entry

emotionalproof (commented):
Great work!

PR #73, opened 2021-02-23
Done

ZacharyDagnall:


README.md
```diff
@@ -89,7 +89,7 @@ We also want to show some of the awesome animal sightings our traveler had. You'
 
 **YOUR NOTES**
 ```
-
+setting the data-id was the hardest (would it have been easier without the dash? not sure.) So i had to use a method called setAttribute(att, val)
 ```
```
BrendaMichelle, line 93: Attributes that begin with `data-` are called data attributes and they give us a standard way to set custom attributes on elements. [Check out this article about them](https://hacks.mozilla.org/2012/10/using-data-attributes-in-javascript-and-css/). I left a comment on your solution as well.

README.md
```diff
@@ -110,5 +110,5 @@ Use Javascript to find the element with the `[data-id='3']` attribute, and *remo
 
 **YOUR NOTES**
 ```
-
+Okay so the query selector needs to be a string and i guess for a non-class non-ID attribute, you have to do '[the thing]' (quotes AND brackets)
 ```
```
BrendaMichelle, line 114: Yea, spot on! Here's some more [CSS selectors](https://www.w3schools.com/cssref/css_selectors.asp)

index.js
```diff
@@ -1,18 +1,33 @@
...
+header.style.color = "blue"
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+document.getElementsByTagName('img')[0].src = traveler.photo
+document.getElementsByTagName('img')[0].alt = "Raffy pic"
+document.getElementsByTagName('h2')[0].innerText = traveler.name
+document.getElementsByTagName('em')[0].innerText = traveler.nickname
+console.log('This is what the traveler object looks like: ', traveler)
```
BrendaMichelle, line 17: Nice! A way to optimize this would be to store the value of `document.getElementsByTagName('img')[0]` in a variable, so that way you can reference the variable on the updates instead of needing JS to search for and find it again.

```js
const img = document.getElementsByTagName('img')[0].src // traversing the DOM _once_ to find the element & storing a reference to it
img.src = traveler.photo
img.alt = "Raffy pic"
```

index.js
```diff
@@ -1,18 +1,33 @@
...
 
 
 /***** Deliverable 4 *****/
-
+const ul = document.querySelector('ul#animals')
```
BrendaMichelle, line 20: 👍

index.js
```diff
@@ -1,18 +1,33 @@
...
+traveler.animalSightings.forEach(function (sighting) {
+    const li = document.createElement('li')
+    li.setAttribute('data-id', `${sighting.id}`)
```
BrendaMichelle, line 23: This works!
- Another solution as you may have read about in the article I linked you to would be:
```js
li.dataset.id = sighting.id
```

You'll also be able to access this property by either writing `li.getAttribute('data-id')` or `li.dataset`

index.js
```diff
@@ -1,18 +1,33 @@
...
+    li.innerHTML = `<p>${sighting.description}</p>
+    <img src="${sighting.photo}" alt="${sighting.species}"/>
+    <a href="${sighting.link}" target="_blank">Here's a video about the ${sighting.species} species!</a>`
+    ul.append(li)
+})
 
 
 /***** Deliverable 5 *****/
+let goAway = document.querySelector('[data-id="3"]')
+goAway.remove()
```
BrendaMichelle, line 33: 💯

BrendaMichelle (commented):
Great work!!

PR #74, opened 2021-02-23
Done

pat-tusza:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```
emotionalproof, line 9: perfect!

index.js
```diff
@@ -1,18 +1,42 @@
...
+const person = document.querySelector("#profile")
+const pic = person.querySelector("img")
+const h2Name = person.querySelector("h2")
+const h2Nick = person.querySelector("em")
 
+pic.src = traveler.photo
+pic.alt = traveler.name
+
+h2Name.innerText = traveler.name
+h2Nick.innerText = traveler.nickname
```
emotionalproof, line 22: great use of variable names

index.js
```diff
@@ -1,18 +1,42 @@
...
+    // console.log(animal)
+    const div = document.createElement("div")
+    div.innerHTML = `<li data-id="${animal.id}">
+                        <p>${animal.description}</p>
+                        <img src="${animal.photo}" alt="${animal.species}"/>
+                        <a href="${animal.link}" target="_blank">Here's a video about the ${animal.species} species!</a>
+                    </li>`
+    const collectionUl = document.querySelector('ul')
+    collectionUl.append(div)
+})
```
emotionalproof, line 36: This is on the correct track and will work fine except the formatting would not be ideal.  Instead of creating 'div', you should create 'li'.  HTML tags have built-in formatting. Putting an li inside of a ul will indent the content and make it easier to read. creating a div for each instance will add extra spacing around each animal sighting card.  Not the end of the world but you can do without it.  Once you create the li, you can do 
li.dataset.id = animal.id
and then 
li.innerHTML =  `<p>${animal.description}</p>
                        ${animal.species}
                        <a href="${animal.link}" target="_blank">Here's a video about the ${animal.species} species!</a>`

Only other note is that the collectionUl variable can be saved above the function since it is not changing and this will save you an extra operation for each instance in animalSightings

index.js
```diff
@@ -1,18 +1,42 @@
...
 
 
 /***** Deliverable 5 *****/
+
+const trash = document.querySelector("[data-id='3']")
+trash.remove()
```
emotionalproof, line 42: Great!

emotionalproof (commented):
Nice work! Thank you for always being vocal on the cohort channels. It will help you become a strong coder and also help the rest of the team

PR #75, opened 2021-02-24
done

mzthomas961:


index.js
```diff
@@ -1,18 +1,52 @@
...
+    <p>${animalSighting.description}</p>
+    <img src="${animalSighting.photo}" alt="${animalSighting.species}"/>
+    <a href="${animalSighting.link}" target="_blank">Here's a video about the ${animalSighting.species} species!</a>
+  </li>`
+    const collection = document.querySelector('ul')
+    collection.append(div) 
+
+
+
+})
```
BrendaMichelle, line 43: Not sure, but looks like you were doing some testing on lines 25 - 38. Otherwise, you don't need lines 25 - 28. You're correctly dynamically creating and appending <li> tags from lines 29 to 43.

```suggestion
/***** Deliverable 4 *****/
const animalSightings = traveler.animalSightings
console.log(animalSightings)

animalSightings.forEach(function (animalSighting){
    const div = document.createElement('div')
    div.classList.add('animalSighting')
   
    div.innerHTML = `<li data-id="${animalSighting.id}">
    <p>${animalSighting.description}</p>
    <img src="${animalSighting.photo}" alt="${animalSighting.species}"/>
    <a href="${animalSighting.link}" target="_blank">Here's a video about the ${animalSighting.species} species!</a>
  </li>`
    const collection = document.querySelector('ul')
    collection.append(div) 



})
```

index.js
```diff
@@ -1,18 +1,52 @@
...
 
 
+const collection = document.querySelector('ul')
+collection.append()
 
 /***** Deliverable 5 *****/
+abc = document.querySelectorAll('[data-id]')
+bca = abc[2]
+bca.remove()
```
BrendaMichelle, line 52: Creative! 
- Make sure you're initializing variables with a declaration keyword and you're using semantic, meaningful variables names. `abc` may be better off as something like `sightingsCollection` or something like that, for example. 
- Also, keep in mind that this solution may not be expandable. For example, imagine now that in another function you get all of the elements with the data-id attribute and you try to remove the element with the data-id value of 4 in the same manner. You'd try removing the element in `abc[3]`, but now that would be the element with the data-id of 5, if it exists.
- Another solution could be:
```js
const elementToRemove = document.querySelectorAll("[data-id='3']")
elementToRemove.remove()
```

BrendaMichelle (commented):
Great work, looks good! Sorry for delay in providing feedback; I failed to notice there were additional submissions for this class earlier.

PR #76, opened 2021-02-24
done

SamFerebee:


index.js
```diff
@@ -1,18 +1,29 @@
...
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
-
-
+console.log('This is what the traveler object looks like: ', traveler)
+document.querySelector("div.traveler > div#profile > h2").textContent = traveler.name;
+document.querySelector("div.traveler > div#profile > em").textContent = traveler.nickname;
+document.querySelector("div.traveler > div#profile > img").src = traveler.photo;
+document.querySelector("div.traveler > div#profile > img").alt = "pic of raffy";
```
emotionalproof, line 15: For a nested querySelector you should be able to omit the ">" and just do "div.traveler div#profile  h2"
you could also save the div#profile to a variable and then run a querySelector on that div
profielDiv.querySelector('h2')

index.js
```diff
@@ -1,18 +1,29 @@
...
 /***** Deliverable 4 *****/
+traveler.animalSightings.forEach(function(obj){
+    const li = document.createElement("li");
+    li.dataset.id = obj.id;
+    li.innerHTML =   `<p>"${obj.description}"</p>
+        <img src="${obj.photo}" alt="${obj.species}"/>
+        <a href="${obj.link}" target="_blank">Here's a video about the ${obj.species} species!</a>`
 
+    document.querySelector("ul#animals").append(li);
+})
```
emotionalproof, line 25: perfect!
you could also do const ul = document.querySelector("ul#animals") above the function and that way you only need to reference it for each iteration, instead of searching for the tag again

index.js
```diff
@@ -1,18 +1,29 @@
...
 
 
 /***** Deliverable 5 *****/
+document.querySelector('[data-id = "3"]').remove();
```
emotionalproof, line 29: great!

emotionalproof (commented):
Nice work Sam!

PR #78, opened 2021-03-16
Update index.js

nicholasdegrate:


index.js
```diff
@@ -1,5 +1,5 @@
 /***** Deliverable 1 *****/
-const header = document.querySelector("#header")
+const header = document.querySelector("h1#header").style.color = "green";
```
lindsayMecher, line 2: nice. Be careful about doing too many things on one line. It would be more readable in 2 lines

```
const header = document.querySelector("h1#header");
header.style.color = "green";
```

This way you have a variable called header which is now reusable throughout the file.

index.js
```diff
@@ -8,11 +8,33 @@ console.log("Here's your header:", header)
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+
+const travelerDiv = document.querySelector('div#profile')
```
lindsayMecher, line 13: great! Love how you saved the entire div to a variable

index.js
```diff
@@ -8,11 +8,33 @@ console.log("Here's your header:", header)
...
+travelerDiv.querySelector('img').src = `${traveler.photo}`
+travelerDiv.querySelector('h2').textContent = `${traveler.name}` 
+travelerDiv.querySelector('em').textContent = `${traveler.nickname}`
```
lindsayMecher, line 16: great work! There are many ways of traversing the DOM, and this works just fine.

You actually do not need to string interpolate the traveler info.
Also make sure to always include an alt attribute for your images.

```
const travelerDiv = document.querySelector('div#profile')
travelerDiv.querySelector('img').src = traveler.photo
travelerDiv.querySelector('img').alt = traveler.name
travelerDiv.querySelector('h2').textContent = traveler.name 
travelerDiv.querySelector('em').textContent = traveler.nickname 
```

index.js
```diff
@@ -8,11 +8,33 @@ console.log("Here's your header:", header)
...
+
+    li.innerHTML = `
+        <p>${obj.description}</p>
+        <img src="${obj.photo}" alt="${obj.species}"/>
+        <a href="${obj.link}" target="_blank">Here's a video about the ${obj.species} species!</a>
+    `
+
+    document.querySelector('ul#animals').append(li)
+
+})
```
lindsayMecher, line 34: Excellent work here! Nice use of creating and appending elements, and also using innerHTML. Creating and appending elements is actually more efficient in terms of time and space complexity, so keep that in mind for optimization in the future. Both ways are perfectly fine, however.

index.js
```diff
@@ -8,11 +8,33 @@ console.log("Here's your header:", header)
...
+
+document.querySelector("[data-id='3']").remove()
```
lindsayMecher, line 36: nice work with this one as well!

lindsayMecher (commented):
Great work, Nick! You seem very comfortable with JS.

PR #86, opened 2021-03-16
completed after review -  Liz

lizburkechi:


index.js
```diff
@@ -1,18 +1,87 @@
...
+
 const header = document.querySelector("#header")
 console.log("Here's your header:", header)
 
 
 /***** Deliverable 2 *****/
 
+header.style.color = "cyan"
+header.style.backgroundColor = "black"
+header.style.border = "chartreuse"
```
BrendaMichelle, line 11: Oooo, nice touches!

index.js
```diff
@@ -1,18 +1,87 @@
...
+
+    <em>${traveler.nickname}</em>
+
+    </div>
+
+    <h4>Great Sightings</h4>
+
+    <ul id="animals">
+    <!-- add an <li> for each animal sighting here -->
+    </ul>`
```
BrendaMichelle, line 36: This works! However, take note that if this code is written after deliverable 4, it would undo deliverable 4's functionality. Also, if we had any references to any of these elements before this deliverable, by using innerHTML we would lose those references. To make the code more scalable in this case, it may be better to search for the elements on the DOM using querySelector since they already exist, store each in a variable, and update the necessary properties of each.

index.js
```diff
@@ -1,18 +1,87 @@
...
+    
+//     animalSighting.dataset.id = sighting.id
+
+//     animalSighting.innerHTML = `<p>${sighting.description}</p>
+
+//     <img src="${sighting.photo}" alt="${sighting.species}"/>
 
+//     <a href="${sighting.link}" target="_blank"> Here's a video about the ${sighting.species} species!</a>
+
+//   </li>`
```
BrendaMichelle, line 55:
```suggestion
```

index.js
```diff
@@ -1,18 +1,87 @@
...
+
+//   listOfSightings.append(animalSighting)
+ 
+// })
```
BrendaMichelle, line 59: This looks good. Only thing that I see may cause some strange display is the closing `</li>` tag on line 55

index.js
```diff
@@ -1,18 +1,87 @@
...
+    li.innerHTML = `<p>${animalSighting.description}</p>
+
+    <img src= "${animalSighting.photo}" alt="${animalSighting.species}"/>
+
+    <a href="${animalSighting.link}" target="_blank">Here's a video about the ${animalSighting.species} species!</a>
+  </li>`
+  
+  console.log(li)
+  ul.append(li)
+})
```
BrendaMichelle, line 80: the combo approach 👏

index.js
```diff
@@ -1,18 +1,87 @@
...
 
 /***** Deliverable 5 *****/
+
+const liToRemove = document.querySelector('[data-id="3"]')
+console.log(liToRemove)
+
+liToRemove.remove()
```
BrendaMichelle, line 87: 👏

BrendaMichelle (commented):
Good work! Left some comments

PR #91, opened 2021-04-06
mini2

chrisgk47:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```
seanpadden, line 9: Great! You can also move your script tag to just above the `</body>` tag to handle this, but we'll ALSO learn a new tool that will help us with this soon.

index.js
```diff
@@ -1,16 +1,23 @@
...
-
+const newHeader = document.querySelector('#header')
+console.log(newHeader.style.color = 'blue')
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
-
-
+console.log('This is what the traveler object looks like: ', traveler)
```
seanpadden, line 12: 👍

index.js
```diff
@@ -1,16 +1,23 @@
...
+const updatedTraveler = document.querySelector('div#profile img')
+updatedTraveler.src = "pics/raffy.jpg"
+updatedTraveler.alt = "Raffy"
+
+const updatedName = document.querySelector('div#profile h2')
+updatedName.textContent = "Chris"
+const updatedName2 = document.querySelector('div#profile em')
+updatedName2.textContent = "Kim"
 /***** Deliverable 4 *****/
```
seanpadden, line 23: **here's a hint w/ deliverable 4**

You need to create [an `li`](https://www.w3schools.com/tags/tag_li.asp) for _each_ animal sighting post and the parent should be the `animalsUl`. You have several options. Think about what we'd need to accomplish this! (hint, `innerHTML` can help you here.)

**Here's a hint w/ deliverable 5**

PR #92, opened 2021-04-06
Done

amautacelt:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```
seanpadden, line 9: Good! We can also move this tag down to just above the `</body>` tag to handle this issue of JS loading before the HTML does.

index.js
```diff
@@ -4,11 +4,22 @@ console.log("Here's your header:", header)
 
 
 /***** Deliverable 2 *****/
-
+const newHeader = document.querySelector("#header")
+console.log(newHeader.style.color = "blue")
```
seanpadden, line 8: 👍

index.js
```diff
@@ -4,11 +4,22 @@ console.log("Here's your header:", header)
...
 
 
 /***** Deliverable 3 *****/
```
seanpadden, line 11: Good!

index.js
```diff
@@ -4,11 +4,22 @@ console.log("Here's your header:", header)
...
+const updatedTravelerName = document.querySelector("div#profile h2")
+updatedTravelerName.textContent = "Carlos"
+
+const updatedTravelerNickname = document.querySelector("div#profile em")
+updatedTravelerNickname.textContent = "Chuck"
+
+console.log('This is what the traveler object looks like: ', traveler)
 
 
 /***** Deliverable 4 *****/
```
seanpadden, line 25: here's some hints w/ deliverable 4
You need to create an li for each animal sighting post and the parent should be the animalsUl. You have several options. Think about what we'd need to accomplish this! (hint, innerHTML can help you here.)

**Here's a hint w/ deliverable 5**

PR #93, opened 2021-04-06
done

m-lawrence:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```
seanpadden, line 9: 👍

index.js
```diff
@@ -5,14 +5,48 @@ console.log("Here's your header:", header)
 
 /***** Deliverable 2 *****/
 
-
+header.style.color = 'purple'
```
seanpadden, line 8: purple!

index.js
```diff
@@ -5,14 +5,48 @@ console.log("Here's your header:", header)
...
+  animalList.append(animalSightingsList)
+}
+
+function allSightings() {
+    traveler.animalSightings.forEach(function(sightingObj) {
+        oneSighting(sightingObj)
+    })
+}
+
+allSightings()
```
seanpadden, line 47: Great!!

index.js
```diff
@@ -5,14 +5,48 @@ console.log("Here's your header:", header)
...
 
 /***** Deliverable 5 *****/
+
+const woopsyAnimal = document.querySelector("[data-id='3']")
+woopsyAnimal.remove()
```
seanpadden, line 52: Great! Check out [this reference](https://www.w3schools.com/cssref/css_selectors.asp) for more CSS selectors

index.js
```diff
@@ -5,14 +5,48 @@ console.log("Here's your header:", header)
...
+
+const firstProfImg = document.querySelector('#profile img')
+firstProfImg.src = "pics/raffy.jpg"
+firstProfImg.alt = "Raffy"
 
+const firstProfName = document.querySelector('div#profile h2')
+firstProfName.innerHTML = "Raffy"
+
+const firstProfNickname = document.querySelector('div#profile em')
+firstProfNickname.innerHTML = "Rafferty"
```
seanpadden, line 21: If you're just changing the text of an element, you should use `textContent` or `innerText` here instead. Ultimately they'll do the same thing, but looking at [this resource](https://stackoverflow.com/questions/35213147/difference-between-textcontent-vs-innertext), which do you think would be better?

index.js
seanpadden, line 47: 💯  Great! Love your use of helper functions here. They'll help down the line if you had to add more elements to our travelers array.

PR #95, opened 2021-04-06
done

JamesHoggardDev:


index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```
seanpadden, line 9: 👍

index.js
```diff
@@ -3,12 +3,15 @@ const header = document.querySelector("#header")
 console.log("Here's your header:", header)
 
 
-/***** Deliverable 2 *****/
 
+/***** Deliverable 2 *****/
+const newHeader = document.querySelector("#header")
+console.log(newHeader.style.color = 'blue')
```
seanpadden, line 9: 🔵  👍

index.js
```diff
@@ -3,12 +3,15 @@ const header = document.querySelector("#header")
...
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+const firstCardImg = document.querySelector('div#profile img')
 
 
 /***** Deliverable 4 *****/
```
seanpadden, line 17: **Here's a hint for deliverable 4**
You need to create an li for each animal sighting post and the parent should be the animalsUl. You have several options. Think about what we'd need to accomplish this! (hint, innerHTML can help you here.)

**Here's a hint for deliverable 5**

index.js
```diff
@@ -3,12 +3,15 @@ const header = document.querySelector("#header")
...
 
+/***** Deliverable 2 *****/
+const newHeader = document.querySelector("#header")
+console.log(newHeader.style.color = 'blue')
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+const firstCardImg = document.querySelector('div#profile img')
```
seanpadden, line 14: Take this step by step. You have the firstCardImg. It's an `image` element. That means it has an `src` tag. How can we access the `src` tag to feed it a URL?

Take this same approach for the other elements and their properties! An `h2` for example, has `textContent`

PR #97, opened 2021-04-06
done.

jabair14:


index.js
```diff
@@ -2,17 +2,44 @@
...
+
+const firstNewImg = document.querySelector("#profile img")
+firstNewImg.src = traveler.photo
+
+firstNewImg.alt = traveler.name 
+
+console.log(firstNewImg)
+
+const travelerName = document.querySelector("#profile > h2")
+travelerName.innerText = traveler.name
```
seanpadden, line 24: Good~! You can also use `textContent` instead of `innerText`. The main difference is that `textContent` doesn't consider the CSS whereas `innerText` does. [Take a look at this stackoverflow post about it](https://stackoverflow.com/questions/35213147/difference-between-textcontent-vs-innertext)

index.js
```diff
@@ -2,17 +2,44 @@
...
+
+
+
+    
 
 
 /***** Deliverable 4 *****/
 
+function renderSightings(animalSightings){
+    const outerLi = document.createElement('li')
```
seanpadden, line 38: This is the right track~! Think about when we need to create these `li` tags. We need to do it **for eachhhhhhhh** element in our `traveler` array. How could we maybe do that??  

Once you're there, you have several options for executing this. Think about what we'd need to accomplish this! (hint, innerHTML can help you here.)

index.js
```diff
@@ -2,17 +2,44 @@
...
+    
+}
+
+// we need to create a li element and then run a forEach loop to populate all of the sightings. Not quite sure how to set it up in this situation however. Deffinitely need to review it first.
```
seanpadden, line 42: yes!

index.js
```diff
@@ -2,17 +2,44 @@
...
 
 
 /***** Deliverable 5 *****/
```
seanpadden, line 45: **Here's a hint for Deliverable 5**

PR #99, opened 2021-04-06
done - worked with matt newton

HenPVel:


data.js
```diff
@@ -1,7 +1,7 @@
 const traveler = {
   id: 1,
-  name: "Raffy",
-  nickname: "Rafferty",
+  name: "Gregster",
+  nickname: "Greg Dwyer",
```
seanpadden, line 4: lol

index.html
```diff
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <link rel="stylesheet" href="styles.css">
   <script src="data.js"></script>
-  <script src="index.js"></script>
+  <script src="index.js" defer></script>
```
seanpadden, line 9: good good. you can also handle this by moving your script tag to just above the `</body>` tag in your HTML file! 

AND we're going to learn another way to handle this soon, too!

index.js
```diff
@@ -1,18 +1,69 @@
 /***** Deliverable 1 *****/
-const header = document.querySelector("#header")
+const header = document.querySelector('h1#header')
 console.log("Here's your header:", header)
+ // used elements and element selector tool to find the correct tag that Amazon Trip text sits in
+// defer the script first in the index.html file
+// use the inspect element tool to find the tag and class i'm searching for
```
seanpadden, line 6: 👍

index.js
```diff
@@ -1,18 +1,69 @@
...
 
-/***** Deliverable 2 *****/
+/***** Deliverable 3 *****/
+console.log('This is what the traveler object looks like: ', traveler)
 
+const travelerName = document.querySelector('div#profile').children[1]
+travelerName.textContent = traveler.name
 
+const travelerNickname = document.querySelector('div#profile').children[2]
+travelerNickname.textContent = traveler.nickname
```
seanpadden, line 20: good use of textContent to account for CSS (as opposed to innerText) [Take a look at this stackoverflow post about it](https://stackoverflow.com/questions/35213147/difference-between-textcontent-vs-innertext)

index.js
```diff
@@ -1,18 +1,69 @@
...
+  sightingsList.append(sightingInstance)
+}
+
+function createAllSightings(array) {
+array.forEach(function(sighting){
+    singleSighting(sighting)
+})
+}
+
+createAllSightings(animalSightingsArray)
```
seanpadden, line 53: Great use of throwing these into helper functions so you can reuse them in the future! Like maybe if we had to add elements to our `travelers` array at some point...

index.js
```diff
@@ -1,18 +1,69 @@
...
+// CREATE A FUNCTION THAT TAKES IN THE CALLBACK AND RUNS IT
+// ON THE sightingsArray
 
 
 /***** Deliverable 5 *****/
+
+
+document.querySelector('ul#animals').children[2].remove()
+
+//HOW DO I GET TO "data-id-3" directly using a method and
```
seanpadden, line 68: You're on the right track and asking the right questions! I wonder if [data](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) could help us?

PR #100, opened 2021-04-06
done

ajnikrant:
Hope it works!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,18 +1,60 @@
 /***** Deliverable 1 *****/
-const header = document.querySelector("#header")
+const header = document.querySelector('#header')
 console.log("Here's your header:", header)
 
 
 /***** Deliverable 2 *****/
-
+header.style.color = "purple"
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+const travelerInfo = document.querySelector('div#profile h2')
+travelerInfo.innerText = traveler.name
+
+const travelerNickname = document.querySelector('div#profile em')
+travelerNickname.innerText = traveler.nickname
+
+const travelerPhoto = document.querySelector('div#profile img')
+travelerPhoto.src = traveler.photo
+travelerPhoto.alt = traveler.name
+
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1,18 +1,60 @@
 /***** Deliverable 1 *****/
-const header = document.querySelector("#header")
+const header = document.querySelector('#header')
 console.log("Here's your header:", header)
 
 
 /***** Deliverable 2 *****/
-
+header.style.color = "purple"
```
seanpadden, line 7: purple!

index.js
```diff
@@ -1,18 +1,60 @@
...
 
 
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+const travelerInfo = document.querySelector('div#profile h2')
+travelerInfo.innerText = traveler.name
+
+const travelerNickname = document.querySelector('div#profile em')
+travelerNickname.innerText = traveler.nickname
```
seanpadden, line 16: 👍  Good! You can also use `textContent` instead of `innerText`. The main difference is that `textContent` doesn't consider the CSS whereas `innerText` does. [Take a look at this stackoverflow post about it](https://stackoverflow.com/questions/35213147/difference-between-textcontent-vs-innertext)

index.js
```diff
@@ -1,18 +1,60 @@
...
+}
+
+function listAllSightings(){
+    const arrayOfSightings = traveler.animalSightings
+    arrayOfSightings.forEach(function(oneSighting) {
+        sightingsLister(oneSighting)
+    })
+}
+
+listAllSightings()
```
seanpadden, line 52: Very nice! Love the helper functions you're using. In a case where we might need to add additional elements to the travelers array, this structure would really help. 💯

index.js
```diff
@@ -1,18 +1,60 @@
...
+
 /***** Deliverable 5 *****/
+
+const thirdLi = document.querySelectorAll(".sighting-container")[4]
```
seanpadden, line 56: What's another way we could find this element, possibly with our `data` element?

PR #101, opened 2021-04-06
Done - Help From Henry Velasquez

majormatt28:


index.js
```diff
@@ -4,15 +4,51 @@ console.log("Here's your header:", header)
...
 /***** Deliverable 3 *****/
-// console.log('This is what the traveler object looks like: ', traveler)
+console.log('This is what the traveler object looks like: ', traveler)
+
+const travelerImageObj = document.querySelector("div#profile").children[0]
+travelerImageObj.src = traveler.photo
+travelerImageObj.alt = 'dog'
+
+const travelerName = document.querySelector("div#profile").children[1]
+travelerName.innerText = traveler.name
```
seanpadden, line 18: 👍  Good! You can also use `textContent` instead of `innerText`. The main difference is that `textContent` doesn't consider the CSS whereas `innerText` does. [Take a look at this stackoverflow post about it](https://stackoverflow.com/questions/35213147/difference-between-textcontent-vs-innertext)

index.js
```diff
@@ -4,15 +4,51 @@ console.log("Here's your header:", header)
...
+  sightingsList.append(sightingInstance)
+}
+
+function createAllSightings() {
+animalSightingsArray.forEach(function(sighting){
+    singleSighting(sighting)
+})
+}
+
+createAllSightings()
```
seanpadden, line 47: Nice use of helper methods here to handle this functionality. If in the future we had to add new elements to our array, having these helper functions would definitely help. 💯

index.js
```diff
@@ -4,15 +4,51 @@ console.log("Here's your header:", header)
...
 
 
 
 /***** Deliverable 5 *****/
+
+const removeUsingDataId = document.querySelector("ul#animals").children[2]
```
seanpadden, line 53: How else could we find this element, maybe with our `data` element?

PR #108, opened 2021-07-21
donee

heyitsbill:


index.js
```diff
@@ -2,17 +2,44 @@
 const header = document.querySelector("#header")
 console.log("Here's your header:", header)
 
+//made js render after body loads
 
 /***** Deliverable 2 *****/
-
+header.style.color="#abcdef"
```
BrendaMichelle, line 8: I like the hex for this color, haha. This is a nice color!

index.js
```diff
@@ -2,17 +2,44 @@
...
+// travelerNicknameNode=document.querySelector("div em")
+// console.log(travelerNameNode)
+//travelerNameNode.innerHTML=traveler.nickname
+//^ doesn't work, em is inline
+
+profileDiv=document.querySelector("#profile")
+profileDiv.innerHTML=`<img src="${traveler.photo}"
+alt="${traveler.name}">
+<h2>${traveler.name}</h2>
+<em>${traveler.nickname}</em>`
```
BrendaMichelle, line 26: This will work here, but be careful using `innerHTML` to update elements because since using innerHTML is essentially recreating those elements, it would remove any existing event listeners as a byproduct.

index.js
```diff
@@ -2,17 +2,44 @@
...
 
 
 /***** Deliverable 3 *****/
 // console.log('This is what the traveler object looks like: ', traveler)
+// travelerNameNode=document.querySelector("h2")
+// travelerNameNode.innerHTML=traveler.name
+
+// travelerNicknameNode=document.querySelector("div em")
+// console.log(travelerNameNode)
+//travelerNameNode.innerHTML=traveler.nickname
```
BrendaMichelle, line 18:
```suggestion
// console.log('This is what the traveler object looks like: ', traveler)
// travelerNameNode = document.querySelector("h2")
// travelerNameNode.textContent = traveler.name

// travelerNicknameNode = document.querySelector("div em")
// console.log(travelerNameNode)
// travelerNameNode.textContent = traveler.nickname
```

index.js
```diff
@@ -2,17 +2,44 @@
...
+
+profileDiv=document.querySelector("#profile")
+profileDiv.innerHTML=`<img src="${traveler.photo}"
+alt="${traveler.name}">
+<h2>${traveler.name}</h2>
+<em>${traveler.nickname}</em>`
 
 
 /***** Deliverable 4 *****/
+var sightings=traveler.animalSightings
```
BrendaMichelle, line 29: Stick with `let` or `const`. Throw `var` away! 🚮

```suggestion
const sightings=traveler.animalSightings
```

index.js
```diff
@@ -2,17 +2,44 @@
...
+let sightingList=document.querySelector("#animals")
+for (let i=0;i<sightings.length;i++){
+    let sighting=sightings[i]
+    item=document.createElement("li")
```
BrendaMichelle, line 33: `const` or `let` 😁  

Random tangent on `const` - There's two schools of thought with `const` in JavaScript. One is to use `const` when you know you are working with values that shouldn't change such as math/physics constants, file versions, etc. Another is to use `const` everywhere and go back and change a declaration to `let` when you come to the point when you need to re-assign that variable.

I tend to follow the second school of thought - I use `const` unless I know I'm going to re-assign a variable later, then I'll use `let`. I do this to eliminate accidental re-assignments.


```suggestion
    const item=document.createElement("li")
```

index.js
```diff
@@ -2,17 +2,44 @@
...
+    <img src="${sighting.photo}" alt="${sighting.species}"/>
+    <a href="${sighting.link}" target="_blank">Here's a video about the ${sighting.species} species!</a>`
+    sightingList.appendChild(item)
+}
 
 
 
 /***** Deliverable 5 *****/
+let badSighting=document.querySelector("li[data-id='3']")
+badSighting.remove()
```
BrendaMichelle, line 45: good!

index.js
BrendaMichelle: https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations

index.js
heyitsbill: ah that makes sense

index.js
heyitsbill: oooh them darn hackers

index.js
heyitsbill: 🤔🤔🤔🤔

PR #109, opened 2021-07-21
assignment done

ekowbe:


README.md
```diff
@@ -47,7 +47,7 @@ Figure out what you need to change to give Javascript access to the `h1#header`
 
 **YOUR NOTES**
 ```
-
+Needed to use window.onload, which makes sure all elements have loaded before getting the header.
 ```
```
BrendaMichelle, line 51: Other options include using the `defer` attribute on the script tag or using `DOMContentLoaded`. Slight variations among all of them, but they will all work here. Here's an interesting comparison between load and DomContentLoaded! http://web.archive.org/web/20150405114023/http://ie.microsoft.com/testdrive/HTML5/DOMContentLoaded/Default.html

README.md
```diff
@@ -57,7 +57,7 @@ Now that you have access to the `h1#header` element, use Javascript to change th
 
 **YOUR NOTES**
 ```
-
+I guess I have to stay in the window.onload function
```
BrendaMichelle, line 60: You can define other functions outside of there, but call them from inside there.

index.js
```diff
@@ -1,18 +1,76 @@
...
+    /***** Deliverable 3 *****/
+    console.log('This is what the traveler object looks like: ', traveler);
+
+    let travelerName = document.body.querySelector("#profile h2");
+    travelerName.innerHTML = `${traveler.name}`;
+
+    let travelerPhoto = document.body.querySelector("#profile img");
+    travelerPhoto.src = `${traveler.photo}`;
+    travelerPhoto.alt = `${traveler.name}`;
+
```
BrendaMichelle, line 21: Good! 
- One recommendation I have is to choose ```textContent``` over ```innerHTML``` when possible especially if you know what you're inserting should be plain text. It's better practice for security reasons. You can read more about the security implications here: https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations
- Another is that you don't have to use [template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) because you're not string interpolating.
```suggestion
    let travelerName = document.body.querySelector("#profile h2");
    travelerName.textContent = traveler.name;

    let travelerPhoto = document.body.querySelector("#profile img");
    travelerPhoto.src = traveler.photo;
    travelerPhoto.alt = traveler.name;

```

index.js
```diff
@@ -1,18 +1,76 @@
...
+    
+
+    let travelerNickName = document.body.querySelector("#profile em");
+    travelerNickName.innerHTML = `${traveler.nickname}`;
```
BrendaMichelle, line 25:
```suggestion
    let travelerNickName = document.body.querySelector("#profile em");
    travelerNickName.textContent = traveler.nickname;
```

index.js
```diff
@@ -1,18 +1,76 @@
...
+        // create a link to a video about the animal
+        let link = document.createElement('a');
+        link.href = `${animal.link}`;
+        link.target = "_blank";
+        link.innerText = `Here's a video about the ${animal.species} species!`
+        li.appendChild(link);
+
+        // render on page
+        ul.appendChild(li);
+
```
BrendaMichelle, line 59: This can be broken out into a helper function (:

Same suggestions here as above relating to innerHTML and string templates.

index.js
```diff
@@ -1,18 +1,76 @@
...
+        li.appendChild(link);
+
+        // render on page
+        ul.appendChild(li);
+        
+    }
+
+    /***** Deliverable 5 *****/
+    let fakeAnimal = document.querySelector("ul#animals li[data-id='3']");
+    fakeAnimal.remove();
```
BrendaMichelle, line 64: good!

